PR #260, opened 2024-08-28
feat(collator): add min_internals_lt calculation in finalize block

serejkaaa512:


SmaGMan:
Blocks GC now relies on diff_tail_len

PR #316, opened 2024-09-17
feat(collator): refactor message groups

serejkaaa512:


SmaGMan:
The logic has been significantly revised in https://github.com/broxus/tycho/pull/531

PR #375, opened 2024-10-24
feat(collator): add check value flow on finalize

serejkaaa512:


0xdeafbeef (approved):
lgtm

SmaGMan:
Should run the check by param in node config. Default should be off.

collator/src/collator/do_collate.rs
```diff
@@ -246,6 +247,7 @@ impl CollatorStdImpl {
                     finalize_params,
                     prepare_groups_wu_total,
                     execute_groups_wu_total,
+                    check_value_flow,
```
SmaGMan, line 250: Should pass full `CollatorConfig` here

SmaGMan (changes requested):
And squash commits, pls.

SmaGMan (approved):

PR #384, opened 2024-10-30
feat(collator): refactor mcdata shards to use short shard description

serejkaaa512:


SmaGMan (changes requested):
Remove duplicating code

PR #397, opened 2024-11-01
fix(consensus): missprint in input buffer

Mododo:
make connfig values more noticeable


consensus/src/engine/input_buffer.rs
```diff
@@ -43,7 +43,8 @@ impl InputBuffer {
 
 impl InputBufferInner for InputBufferData {
     fn push(&mut self, ext_in_msg: Bytes) {
-        if self.buffer_bytes == 0 || self.data_bytes == 0 {
```
Mododo, line 46: mf `data_bytes` missprint

Rexagon (approved):

SmaGMan (approved):

PR #412, opened 2024-11-04
Do not remove persistent blocks for shards

Rexagon:


storage/src/store/block_handle/handle.rs
```diff
@@ -72,6 +72,10 @@ impl BlockHandle {
         self.inner.meta.flags().contains(BlockFlags::IS_APPLIED)
     }
 
+    pub fn is_persistent(&self) -> bool {
+        self.inner.meta.flags().contains(BlockFlags::IS_PERSISTENT) || self.inner.id.seqno == 0
```
0xdeafbeef, line 76: Where can we get blocks with 0 seqno?

0xdeafbeef (approved):
lgtm

storage/src/store/block_handle/handle.rs
Rexagon: Block handles are created even for zerostates

SmaGMan (approved):

PR #415, opened 2024-11-05
feat(collator): remove async where it is not needed

serejkaaa512:


collator/src/manager/mod.rs
```diff
@@ -136,26 +136,26 @@ where
...
 
         let state_cloned = state.clone();
-        self.spawn_task(method_to_async_closure!(
-            detect_top_processed_to_anchor_and_notify_mempool,
-            true,
-            state_cloned,
-            processed_upto
-        ))
-        .await?;
+        self.spawn_task_blocking(move |worker| {
```
0xdeafbeef, line 145: `future::ready`

PR #416, opened 2024-11-05
Cycle block provider in block strider

pashinov:


collator/tests/collation_tests.rs
```diff
@@ -33,6 +33,8 @@ impl BlockProvider for StrangeBlockProvider {
     type GetNextBlockFut<'a> = BoxFuture<'a, OptionalBlockStuff>;
     type GetBlockFut<'a> = BoxFuture<'a, OptionalBlockStuff>;
 
+    type ResetFut<'a> = BoxFuture<'a, ()>;
```
0xdeafbeef, line 36: Change to https://doc.rust-lang.org/std/future/struct.Ready.html

core/src/block_strider/provider/blockchain_provider.rs
```diff
@@ -197,6 +197,7 @@ impl BlockchainBlockProvider {
 impl BlockProvider for BlockchainBlockProvider {
     type GetNextBlockFut<'a> = BoxFuture<'a, OptionalBlockStuff>;
     type GetBlockFut<'a> = BoxFuture<'a, OptionalBlockStuff>;
+    type ResetFut<'a> = BoxFuture<'a, ()>;
```
0xdeafbeef, line 200: same

core/src/block_strider/provider/mod.rs
```diff
@@ -126,6 +162,77 @@ impl<T1: BlockProvider, T2: BlockProvider> BlockProvider for ChainBlockProvider<
...
+                self.left.reset().await;
+                self.left.get_next_block(prev_block_id).await
+            } else {
+                self.right.reset().await;
+                self.right.get_next_block(prev_block_id).await
+            };
+
+            // Wait for timeout if a new provider also didn't return value
+            if res.is_none() {
+                tokio::time::sleep(Duration::from_secs(60)).await;
```
0xdeafbeef, line 210: just return `None`; caller should retry with it's own intervals

0xdeafbeef (approved):
We can move constants to the `CycleBlockProvider` constructor, but a 10 to 1 ratio is ok for me

cli/src/node/mod.rs
```diff
@@ -434,10 +434,10 @@ impl Node {
             .with_provider(
                 collator
                     .new_sync_point(CollatorSyncContext::Historical)
-                    .chain(archive_block_provider)
+                    .chain(archive_block_provider.clone())
                     .chain(collator.new_sync_point(CollatorSyncContext::Recent))
                     .chain((
-                        blockchain_block_provider,
+                        blockchain_block_provider.cycle(archive_block_provider),
```
Rexagon, line 440: `ArchiveBlockProvider` must reset its own state before it is going to return `None` on `get_next_block`

core/src/block_strider/provider/mod.rs
```diff
@@ -128,6 +151,64 @@ impl<T1: BlockProvider, T2: BlockProvider> BlockProvider for ChainBlockProvider<
...
+
+        let res = if !is_right {
+            $self.left.$method($param).await
+        } else {
+            $self.right.$method($param).await
+        };
+
+        if res.is_some() {
+            // Reset retry counter when successes
+            $self.retry_counter.store(0, Ordering::Release);
```
Rexagon, line 166: Retry could be implemented via a separate provider.

```rust
blockchain_block_provider.retry(10).cycle(archive_block_provider.retry(1))
```

core/src/block_strider/provider/blockchain_provider.rs
```diff
@@ -93,12 +93,12 @@ impl BlockchainBlockProvider {
                             return parsed;
                         }
                     }
-                    None => tracing::warn!(?prev_block_id, "block not found"),
+                    None => {
+                        tracing::warn!(?prev_block_id, "block not found");
```
0xdeafbeef, line 97: Just return `None`. We should probably log inside the retry provider when it has exceeded its attempts. Otherwise, it creates useless noise

core/src/block_strider/provider/blockchain_provider.rs
```diff
@@ -130,12 +130,12 @@ impl BlockchainBlockProvider {
                             return parsed;
                         }
                     }
-                    None => tracing::warn!(%block_id, "block not found"),
+                    None => {
+                        tracing::warn!(%block_id, "block not found");
```
0xdeafbeef, line 134: same

0xdeafbeef (changes requested):

0xdeafbeef (approved):
lgtm

network/src/network/connection_manager.rs
```diff
@@ -326,9 +327,8 @@ impl ConnectionManager {
                 });
                 metrics::gauge!(METRIC_CONNECTIONS_PARTIAL).increment(1);
             }
-            Into0RttResult::InvalidConnection(e) => {
-                // TODO: Lower log level to trace/debug?
-                tracing::warn!(%remote_addr, "invalid incoming connection: {e}");
```
0xdeafbeef, line 331: 've you ever seen this logs? Maybe it's ok?

Rexagon: We have not yet spammed with invalid peer ids (malformed ed25519 pubkeys). It is the case when this error is visible, and we can just ignore that spam (no need to flood logs with it)

PR #429, opened 2024-11-07
Feature: handle mempool paused

Mododo:


collator/src/mempool/impls/std_impl/cache.rs
```diff
@@ -55,20 +69,23 @@ impl Cache {
...
+                        match data.anchors.get(&anchor_id) {
+                            Some(found) => return Some(found.clone()),
+                            None => {
+                                tracing::warn!(
+                                    target: tracing_targets::MEMPOOL_ADAPTER,
+                                    %anchor_id,
+                                    is_paused = Some(data.is_paused).filter(|x| *x),
+                                    "Anchor is unknown, waiting"
+                                );
+                            }
```
Mododo, line 87: looks like this is the only change in ``Cache` behavior: if mempool is expected to be not paused, then collator will wait for anchor to be received.

collator/src/mempool/impls/std_impl.rs
```diff
@@ -377,16 +393,41 @@ impl MempoolAdapter for MempoolAdapterStdImpl {
...
+
+        let result = match self
+            .expect_running(top_processed_to_anchor, anchor_id)
+            .await
+        {
+            Some(true) => match self.cache.get_anchor_by_id(anchor_id).await {
+                Some(anchor) => GetAnchorResult::Exist(anchor),
+                None => GetAnchorResult::NotExist,
+            },
+            Some(false) | None => GetAnchorResult::MempoolPaused, // also paused until init
```
Mododo, line 409: have doubts on what to return on `None`. It's the case when `Engine` is not run yet , but anchors are already requested.

Mododo:
marking as draft request because would like to squash all my adapter changes into single commit and rebase other changes

collator/src/mempool/impls/std_impl.rs
SmaGMan: How it worked before when the `expect_running()` check was not present?

Mododo: It returned `None` immediately, even in case when anchor could be received (or not) after some time. That behavior was inherited from old versions when mempool was unable to sync (regardless pause mode).

Mododo: it returned `NotExist` (`None`), but anchor may be received after mempool is run.

PR #435, opened 2024-11-08
feat(collator): impl validate params

serejkaaa512:


collator/src/collator/do_collate.rs
```diff
@@ -246,6 +247,7 @@ impl CollatorStdImpl {
                     finalize_params,
                     prepare_groups_wu_total,
                     execute_groups_wu_total,
+                    validate_config,
```
SmaGMan, line 250: Let's pass full `CollatorConfig` here.

collator/src/collator/build_block.rs
```diff
@@ -555,12 +557,13 @@ impl CollatorStdImpl {
         config_params: Option<BlockchainConfig>,
         prev_state: &ShardStateStuff,
         prev_processed_to_anchor: u32,
+        validate_config: bool,
```
SmaGMan, line 560: Let's pass full `CollatorConfig` here.

block-util/src/config/mod.rs
```diff
@@ -66,15 +62,13 @@ pub trait BlockchainConfigExt {
...
-        _mandatory_params: Option<Dict<u32, ()>>,
-    ) -> Result<bool> {
-        // TODO: refer to https://github.com/everx-labs/ever-block/blob/master/src/config_params.rs#L452
-        // STUB: currently should not be invoked in prototype
-        // todo!()
-        Ok(true)
+    fn validate_params(&self) -> bool {
+        match self.get::<ConfigParam0>() {
+            Ok(Some(config_address)) if config_address == self.address => {
+                self.get_mandatory_params().is_ok() && self.get_critical_params().is_ok()
```
SmaGMan, line 68: In this case we should get the list of mandatory params and check that config contains this params and they can be parsed. We should merge lists of mandatory and critical params and check them all.

SmaGMan (changes requested):
Needs to fix params validation

SmaGMan:
And squash commits before merge.

collator/src/collator/build_block.rs
```diff
@@ -555,12 +557,13 @@ impl CollatorStdImpl {
...
         prev_state: &ShardStateStuff,
         prev_processed_to_anchor: u32,
+        collator_config: Arc<CollatorConfig>,
     ) -> Result<(McStateExtra, u32)> {
         // 1. update config params and detect key block
         let prev_state_extra = prev_state.state_extra()?;
         let prev_config = &prev_state_extra.config;
         let (config, mut is_key_block) = if let Some(new_config) = config_params {
-            if !new_config.validate_params(true, None)? {
+            if collator_config.validate_config && !new_config.validate_params() {
```
Rexagon, line 566: Check only if changed (move after `is_key_block`)

block-util/src/config/mod.rs
```diff
@@ -66,15 +62,43 @@ pub trait BlockchainConfigExt {
...
-        &self,
-        _relax_par0: bool,
-        _mandatory_params: Option<Dict<u32, ()>>,
-    ) -> Result<bool> {
-        // TODO: refer to https://github.com/everx-labs/ever-block/blob/master/src/config_params.rs#L452
-        // STUB: currently should not be invoked in prototype
-        // todo!()
-        Ok(true)
+    fn validate_params(&self) -> bool {
+        match self.get::<ConfigParam0>() {
```
Rexagon, line 66: It can be greatly simplified with `let .. else`.

And it might be useful to return `Result<()>` instead of just bool to know what is wrong with the config.

Rexagon (changes requested):

block-util/src/config/mod.rs
Rexagon: Critical params are just some params that have a special requirements for update proposals. They can be absent

PR #438, opened 2024-11-09
Feature/handle mempool paused sq

SmaGMan:


collator/src/collator/mod.rs
```diff
@@ -916,46 +933,61 @@ impl CollatorStdImpl {
             .get_last_imported_anchor_id_and_ct()
             .unwrap_or_default();
 
+        // do not import anchor if mempool may be paused
+        // needs to process more anchors in collator first
+        if id - top_processed_to_anchor > max_consensus_lag_rounds / 2 {
```
Mododo, line 938: no possibility for overflow here?  
for example, if we collated, then synced, then began collating again, and no new anchors were imported from mempool (local mempool disconnected or is syncing) 
maybe `saturating_sub`?

consensus/src/engine/impl_.rs
```diff
@@ -285,6 +291,11 @@ impl Engine {
                 continue;
             }
 
+            if is_paused {
+                is_paused = false;
+                self.committed_info_tx.send(MempoolOutput::Running).ok();
```
0xdeafbeef, line 296: Won't unwrapping here be better? Listener must be present; otherwise it can become a silent bug

0xdeafbeef (approved):
lgtm

consensus/src/engine/impl_.rs
Mododo: it's to shut down gracefully.
warning on channel close is logged in adapter with `scopeguard`, when consumer exits loop in spawned task

Mododo (approved):

collator/src/collator/mq_iterator_adapter.rs
```diff
@@ -108,209 +115,146 @@ impl<V: InternalMessageValue> QueueIteratorAdapter<V> {
...
+                self.iterator_opt = Some(iterator);
+
+                true
             }
+            (false, true, InitIteratorMode::OmitNextRange) => false,
+            _ => {
+                // when current iterator exists or existing ranges fully read
+                // then try to calc new ranges from current states
+                // add masterchain default range if not exist
+                let mc_shard_id = ShardIdent::new_full(-1);
```
Rexagon, line 173: `ShardIdent::MASTERCHAIN`

Rexagon (approved):

PR #441, opened 2024-11-11
perf: move states gc to spawn_blocking

0xdeafbeef:


storage/src/store/shard_state/mod.rs
```diff
@@ -211,14 +209,26 @@ impl ShardStateStorage {
...
-                batch.delete_cf(&shard_states_cf.bound(), key);
-                raw.write_opt(batch, cells_write_options)?;
+
+                let db = self.db.clone();
+                let cell_storage = self.cell_storage.clone();
+                let key = key.to_vec();
+                let mut batch = rocksdb::WriteBatch::default();
+
+                let total = tokio::task::spawn_blocking(move || {
+                    let total = cell_storage.remove_cell(&mut batch, &root_hash)?;
```
Rexagon, line 222: Please reuse `alloc` here. You can just move it into and out of the spawn since it is `Send`

0xdeafbeef: done

Rexagon (approved):

pashinov (approved):

PR #445, opened 2024-11-11
feat(collator): collate empty shard block by timeout when no messages

SmaGMan:
Waiting when macros in `everscale-types` will be updated to support versioning. Then will update `CollationConfig` in types with backward compatibility and finish this feature.

### DIFF
--- a/collator/src/collator/mod.rs
+++ b/collator/src/collator/mod.rs
@@ -528,6 +528,7 @@ impl CollatorStdImpl {
 
             // update mc_data if newer
             if working_state.mc_data.block_id.seqno < mc_data.block_id.seqno {
+                working_state.collation_config = Arc::new(mc_data.config.get_collation_config()?);
                 working_state.mc_data = mc_data;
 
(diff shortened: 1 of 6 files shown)


collator/src/collator/mod.rs
```diff
@@ -1547,21 +1549,41 @@ impl CollatorStdImpl {
...
+            .unwrap();
+        let prev_block_chain_time = working_state.prev_shard_data_ref().gen_chain_time();
+
+        // when no messages for processing
+        // then force empty shard block collation if related timeout elapsed
+        if matches!(
+            try_collate_check,
+            TryCollateCheck::ImportedAnchorHasNoExternals
+        ) {
+            let empty_sc_block_interval_ms =
```
drmick, line 1564: Perhaps using `Duration` would be  abetter approach here.

drmick (approved):

collator/src/collator/mod.rs
Rexagon: We can't use Duration in blockchain configs due to different requirements of units for different params

PR #450, opened 2024-11-12
fix(cli): compute elections timeline for the current time

Rexagon:


pashinov (approved):

cli/src/cmd/elect/mod.rs
```diff
@@ -192,10 +207,14 @@ impl CmdRun {
                 .get_elector_data(&config.elector_addr)
                 .await
                 .context("failed to get elector data")?;
+
+            let now = now_sec();
+            let time_diff = now.saturating_sub(gen_utime).min(MAX_TIME_DIFF);
             let timeline = config
-                .compute_elections_timeline(gen_utime)
+                .compute_elections_timeline(now)
```
0xdeafbeef, line 214: Won't using `now` here instead of blockchain time break something?

Rexagon: This is an important part of the fix. If everything is ok and elections logic was not shifted we should decide whether elections are running based on the current time to get absolute timestamps

PR #480, opened 2024-11-29
feature(collator): queue diffs tail

drmick:


collator/src/collator/do_collate/finalize.rs
```diff
@@ -97,6 +108,8 @@ impl Phase<FinalizeState> {
         )
         .serialize();
 
+        let processed_upto = diff_with_messages.processed_upto.clone();
```
SmaGMan, line 111: Let's rename this into `diff_processed_to` across collation methods to separate from real `processed_upto { processed_to, read_to }`.

collator/src/collator/do_collate/finalize.rs
```diff
@@ -127,29 +145,43 @@ impl Phase<FinalizeState> {
...
         let wu_params_finalize = self
             .state
             .collation_config
             .work_units_params
             .finalize
             .clone();
 
+        let mut processed_upto = FastHashMap::default();
+
+        for (shard_ident, processed_upto_stuff) in
```
SmaGMan, line 179: Seems like it will be better to use `iter().map().collect()`

collator/src/collator/types.rs
```diff
@@ -1277,6 +1279,7 @@ pub struct UpdateQueueDiffResult {
     pub has_unprocessed_messages: bool,
     pub diff_messages_len: usize,
     pub create_queue_diff_elapsed: Duration,
+    pub processed_upto: BTreeMap<ShardIdent, QueueKey>,
```
SmaGMan, line 1282: Let's rename into `processed_to`

collator/src/collator/do_collate/finalize.rs
```diff
@@ -554,6 +589,36 @@ impl Phase<FinalizeState> {
...
+
+                    // Extract processed information for updated shards
+                    let processed_upto = self
+                        .state
+                        .collation_data
+                        .top_shard_blocks
+                        .iter()
+                        .find(|top_block_info| top_block_info.block_id.shard == *shard_id)
+                        .map(|top_block_info| {
+                            FastHashMap::from(top_block_info.processed_upto.clone())
```
SmaGMan, line 607: Can we just clone here?

collator/src/collator/do_collate/finalize.rs
```diff
@@ -128,29 +146,46 @@ impl Phase<FinalizeState> {
...
+        } = ctx;
+
         let wu_params_finalize = self
             .state
             .collation_config
             .work_units_params
             .finalize
             .clone();
 
+        let processed_to = self
```
SmaGMan, line 178: Why not to take just from `queue_diff`?

collator/src/collator/do_collate/finalize.rs
```diff
@@ -555,6 +593,34 @@ impl Phase<FinalizeState> {
                     self.state.collation_data.processed_upto.externals.as_ref(),
                 );
 
+                let mut shards_processed_to = FastHashMap::default();
+
+                for (shard_id, shard_data) in shards.iter() {
+                    if !shard_data.top_sc_block_updated {
```
SmaGMan, line 599: Looks like we will totally skip shards that were not updated instead of reading info from the previous `McData`.

collator/src/collator/do_collate/finalize.rs
```diff
@@ -587,7 +653,7 @@ impl Phase<FinalizeState> {
             block: new_block,
             is_key_block: new_block_info.key_block,
             prev_blocks_ids: self.state.prev_shard_data.blocks_ids().clone(),
-            top_shard_blocks_ids: self.state.collation_data.top_shard_blocks_ids.clone(),
+            top_shard_blocks: self.state.collation_data.top_shard_blocks.clone(),
```
SmaGMan, line 656: Why do we need to change this to the `Vec<TopShardBlockInfo>` if further we use only blocks ids [here](https://github.com/broxus/tycho/pull/480/files#diff-e1c4abb5972d10b9f9c9a8593a763efd50a52d5c9dc1cbaf000ba392260cb052R128)

collator/src/collator/do_collate/finalize.rs
SmaGMan: I suggest to write a helper function and make a unit test.

block-util/src/queue/queue_diff.rs
```diff
@@ -27,11 +27,11 @@ impl QueueDiffStuffBuilder {
     }
 
     // TODO: Use iterator of `(ShardIdent, QueueKey)`?
-    pub fn with_processed_upto<'a, I>(mut self, processed_upto: I) -> Self
+    pub fn with_processed_to<'a, I>(mut self, processed_to: I) -> Self
     where
         I: IntoIterator<Item = (ShardIdent, u64, &'a HashBytes)>,
     {
-        self.inner_mut().diff.processed_upto = processed_upto
+        self.inner_mut().diff.processed_upto = processed_to
```
SmaGMan, line 34: Let's rename `QueueDiff.processed_upto` into `QueueDiff.processed_to` as well. But we cannot update `proto.tl` right now because type id could be changed. So we should place a `TODO` comment to the field:
```
// TODO: should rename field in `proto.tl` on network reset
```

Let's rename `mod processed_upto_map` into `mod processed_to_map`.

collator/src/queue_adapter.rs
```diff
@@ -99,11 +105,12 @@ impl<V: InternalMessageValue> MessageQueueAdapter<V> for MessageQueueAdapterStdI
         diff: QueueDiffWithMessages<V>,
         block_id_short: BlockIdShort,
         hash: &HashBytes,
+        end_key: QueueKey,
     ) -> Result<()> {
         let time = std::time::Instant::now();
         let len = diff.messages.len();
-        let processed_upto = diff.processed_upto.clone();
-        self.queue.apply_diff(diff, block_id_short, hash)?;
+        let processed_upto = diff.processed_to.clone();
```
SmaGMan, line 112: Rename variable `processed_upto` into `processed_to`.

collator/src/collator/do_collate/mod.rs
```diff
@@ -278,25 +280,80 @@ impl CollatorStdImpl {
...
 
         let finalize_block_timer = std::time::Instant::now();
 
+        let mut min_processed_to: Option<QueueKey> = None;
+
+        // Get current and masterchain processed values
+        let current_processed_to = processed_to.get(&shard_id).cloned();
+        let mc_processed_to = mc_data.processed_upto.internals.get(&shard_id).cloned();
+
+        if shard_id.is_masterchain() {
```
SmaGMan, line 296: Let's make a helper function and write some unit tests.

collator/src/collator/do_collate/mod.rs
```diff
@@ -278,25 +280,80 @@ impl CollatorStdImpl {
...
-        ) = finalize_phase.update_queue_diff(mq_iterator_adapter, shard_id, mq_adapter)?;
+        ) = finalize_phase.update_queue_diff(mq_iterator_adapter, shard_id, mq_adapter.clone())?;
 
         let finalize_block_timer = std::time::Instant::now();
 
+        let mut min_processed_to: Option<QueueKey> = None;
+
+        // Get current and masterchain processed values
+        let current_processed_to = processed_to.get(&shard_id).cloned();
+        let mc_processed_to = mc_data.processed_upto.internals.get(&shard_id).cloned();
```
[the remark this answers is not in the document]
SmaGMan: Or may get value in place further [here](https://github.com/broxus/tycho/pull/480/files#diff-9164155323991cf162b6465dec02022365f7b8ac572dee1d7142af656ab574a8R331) without additional temporary variables.

collator/src/internal_queue/queue.rs
```diff
@@ -268,8 +286,35 @@ where
         Ok(())
     }
 
-    fn clear_session_state(&self) -> Result<()> {
-        self.diffs.clear();
-        self.session_state.truncate()
+    fn clear_uncommitted_state(&self) -> Result<()> {
```
SmaGMan, line 289: We call this method when block mismatched. Then we run sync to applied mc block and apply required queue diff for this mc block. Possibly we can get incorrect `diff_tail_len` because we keep `committed_diffs` unchanged. Should check this for all places were `clear_uncommitted_state` called.

collator/src/manager/blocks_cache.rs
```diff
@@ -149,33 +155,43 @@ impl BlocksCache {
                     mc_block_key,
                 )
             };
+
+            let processed_to = mc_block_entry
```
SmaGMan, line 159: Possibly we can use the same map type for `processed_to` across all the code without any perf cons? The conversion from `BTreeMap` to `HasMap` and vice versa for this data looks wired.

core/src/block_strider/subscriber/mod.rs
```diff
@@ -420,3 +421,59 @@ pub mod test {
...
+    let mut min_tail_len = None;
+
+    for block_handle in shard_block_handles {
+        let block = storage
+            .block_storage()
+            .load_block_data(&block_handle)
+            .await?;
+        let tail_len = block.block().out_msg_queue_updates.tail_len as usize;
+
+        min_tail_len = Some(min_tail_len.map_or(tail_len, |min: usize| min.min(tail_len)));
```
SmaGMan, line 439: Why do we get minimal tail length from shards? Should it be the longest tail instead?

SmaGMan (approved):

SmaGMan:
Will be included into [feat/msgs-proc-logic-v3](https://github.com/broxus/tycho/tree/feat/msgs-proc-logic-v3)

PR #486, opened 2024-12-04
feat(collator): use prev data accounts extra total count for finalize…

serejkaaa512:
… wu calculation


collator/src/collator/do_collate/finalize.rs
```diff
@@ -344,19 +344,27 @@ impl Phase<FinalizeState> {
             );
 
             let accounts_count = self.state.collation_data.accounts_count;
```
SmaGMan, line 346: Needs to rename `account_count` into `updated_accounts_count` to avoid confusion. We might mistakenly assume that this is new total accounts count in the state.

PR #491, opened 2024-12-05
feat(collator): inline more data into `ShardAccountStuff`

Rexagon:


collator/src/collator/types.rs
```diff
@@ -567,35 +568,50 @@ pub(super) struct ShardAccountStuff {
...
+                ..
+            }) = account.state
+            {
+                libraries = acc_libs;
+                special = acc_flags.unwrap_or_default();
+            }
+            balance = account.balance;
+            exists = true;
+        } else {
+            libraries = Dict::new();
```
SmaGMan, line 600: We already assigned these values to `libraries` and `special` on declaration.

SmaGMan (approved):

serejkaaa512 (approved):

PR #498, opened 2024-12-09
feat: add cells cache metrics

0xdeafbeef:


Rexagon:
Included in https://github.com/broxus/tycho/pull/500

PR #510, opened 2024-12-27
Refactor/genesis info

Mododo:
[diff for everscale-types](https://github.com/broxus/everscale-types/compare/2d8c5f08fd713fcea3e1d6d0720b7a691742f5ee...488e42786c84dcf66a329d446a7612d0636096b6)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -136,6 +136,7 @@ tycho-util = { path = "./util", version = "0.1.5" }
 
 [patch.crates-io]
 weedb = { version = "0.3.8", git = "https://github.com/broxus/weedb.git", rev = "59728ea0c8703dd28a4c37dee05c1321cd81b966" }
+everscale-types = { git = "https://github.com/broxus/everscale-types.git", rev = "488e42786c84dcf66a329d446a7612d0636096b6" }
 
 [workspace.lints.rust]
 future_incompatible = "warn"
--- a/collator/src/collator/do_collate/finalize.rs
+++ b/collator/src/collator/do_collate/finalize.rs
@@ -719,11 +719,8 @@ impl Phase<FinalizeState> {
         let mut consensus_info = prev_state_extra.consensus_info;
 
         if let Some(mp_cfg_override) = &collation_data.mempool_config_override {
-            if mp_cfg_override.start_round >= consensus_info.genesis_round
-                && mp_cfg_override.genesis_time_millis > consensus_info.genesis_millis
-            {
-                consensus_info.genesis_round = mp_cfg_override.start_round;
-                consensus_info.genesis_millis = mp_cfg_override.genesis_time_millis;
+            if (mp_cfg_override.genesis_info).overrides(&consensus_info.genesis_info) {
+                consensus_info.genesis_info = mp_cfg_override.genesis_info;
 
                 is_key_block = true;
             }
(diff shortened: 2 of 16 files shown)


Cargo.toml
```diff
@@ -136,6 +136,7 @@ tycho-util = { path = "./util", version = "0.1.5" }
 
 [patch.crates-io]
 weedb = { version = "0.3.8", git = "https://github.com/broxus/weedb.git", rev = "59728ea0c8703dd28a4c37dee05c1321cd81b966" }
+everscale-types = { git = "https://github.com/broxus/everscale-types.git" }
```
Rexagon, line 139: Pin a specific commit as a revision for types

collator/src/collator/do_collate/finalize.rs
```diff
@@ -719,11 +719,10 @@ impl Phase<FinalizeState> {
         let mut consensus_info = prev_state_extra.consensus_info;
 
         if let Some(mp_cfg_override) = &collation_data.mempool_config_override {
-            if mp_cfg_override.start_round >= consensus_info.genesis_round
-                && mp_cfg_override.genesis_time_millis > consensus_info.genesis_millis
+            if mp_cfg_override.genesis_info.start_round >= consensus_info.genesis_info.start_round
```
Rexagon, line 722: Maybe impl an `Ord`/`PartialOrd` trait for `GenesisInfo`? Or at least a single method to determine that info is "after"

collator/src/manager/mod.rs
```diff
@@ -1168,14 +1168,15 @@ where
             let last_consesus_info = self
                 .blocks_cache
                 .get_consensus_info_for_mc_block(&last_applied_mc_block_key)?;
-            if mp_cfg_override.start_round >= last_consesus_info.genesis_round
-                && mp_cfg_override.genesis_time_millis > last_consesus_info.genesis_millis
+            if mp_cfg_override.genesis_info.start_round
```
Rexagon, line 1171: ^ see cmp comment

collator/src/mempool/impls/std_impl.rs
```diff
@@ -283,8 +283,8 @@ impl MempoolAdapter for MempoolAdapterStdImpl {
             );
 
             if let Some(genesis) = (config_guard.builder.get_genesis()).filter(|genesis| {
-                genesis.start_round >= new_cx.consensus_info.genesis_round
-                    && genesis.time_millis > new_cx.consensus_info.genesis_millis
+                genesis.start_round >= new_cx.consensus_info.genesis_info.start_round
```
Rexagon, line 286: ^ see cmp comment

Rexagon (changes requested):

collator/src/collator/do_collate/finalize.rs
Mododo: unfortunately, would like to have comparison in tycho repo so it could be easily changed in place, while need to have the struct in types to be accessible across crates

Cargo.toml
Mododo: the latest commit in this PR (including this line and the one in `Cargo.lock`) should be removed, when the new version of types is ready. or do we want to use the patch?

Mododo: done; patch with revision to merge this commit earlier than next types version

collator/src/collator/do_collate/finalize.rs
Mododo: done; added a single responsibility method

Rexagon (approved):

PR #526, opened 2025-01-08
fix(collator): trim tail fix

drmick:


collator/src/collator/do_collate/finalize.rs
```diff
@@ -115,6 +115,7 @@ impl Phase<FinalizeState> {
 
         let queue_diff_messages_count = queue_diff_with_msgs.messages.len();
 
+        let end_key = QueueKey::max_for_lt(self.state.collation_data.next_lt);
```
SmaGMan, line 118: Why not to take already existing `max_message`?

collator/src/manager/blocks_cache.rs
```diff
@@ -949,7 +951,7 @@ struct ReceivedBlockContext {
     queue_diff: QueueDiffStuff,
     out_msgs: Lazy<OutMsgDescr>,
     ref_by_mc_seqno: u32,
-    // NOTE: `BlockStuff` can also be added here if needed since it's already loaded
+    end_lt: u64, // NOTE: `BlockStuff` can also be added here if needed since it's already loaded
```
SmaGMan, line 954: Please move comment to a separate line.

collator/src/manager/blocks_cache.rs
```diff
@@ -1002,6 +1007,7 @@ impl ReceivedBlockContext {
             queue_diff,
             out_msgs,
             ref_by_mc_seqno,
+            end_lt,
```
SmaGMan, line 1010: Seems like we can just use `queue_diff.as_ref().max_message` instead.

collator/src/manager/mod.rs
```diff
@@ -1289,12 +1292,15 @@ where
...
 
+                    let end_lt = block_stuff.block().info.load()?.end_lt;
+
                     let queue_diff_with_messages =
                         QueueDiffWithMessages::from_queue_diff(&queue_diff_stuff, &out_msgs)?;
                     prev_queue_diffs.push((
                         queue_diff_with_messages,
                         *queue_diff_stuff.diff_hash(),
                         prev_block_id,
+                        end_lt,
```
SmaGMan, line 1303: Seems like we can just use `queue_diff_stuff.as_ref().max_message` instead.

collator/src/manager/mod.rs
```diff
@@ -1307,10 +1313,15 @@ where
         }
 
         // apply required previous queue diffs
-        while let Some((diff, diff_hash, block_id)) = prev_queue_diffs.pop() {
+        while let Some((diff, diff_hash, block_id, end_lt)) = prev_queue_diffs.pop() {
```
SmaGMan, line 1316: Here we can get `max_message: QueueKey` instead of only LT.

collator/src/manager/types.rs
```diff
@@ -162,6 +162,8 @@ pub(super) enum BlockCacheEntryData {
 
         /// Additional shard block cache info
         additional_shard_block_cache_info: Option<AdditionalShardBlockCacheInfo>,
+        /// End lt of the block
+        end_lt: u64,
```
SmaGMan, line 166: Seems like we do not need this field. Refer to prev comments.

PR #529, opened 2025-01-09
ci: add a script to set recommended sysctl settings

0xdeafbeef:


scripts/set-sysctl.sh
```diff
@@ -0,0 +1,8 @@
+set -e
```
Rexagon, line 1: Set `x` flag since this script will be used from `sudo` and it might be a good idea to show what is going on

Rexagon, line 1: Add `#!/usr/bin/env bash` to be consistent with other scripts

Rexagon (changes requested):

Rexagon (approved):

pashinov (approved):

PR #531, opened 2025-01-09
messages processing logic v3

SmaGMan:


0xdeafbeef (approved):

block-util/src/block/block_stuff.rs
```diff
@@ -24,7 +24,7 @@ impl BlockStuff {
     pub const BOOT_OFFSET: Duration = Duration::from_secs(12 * 3600);
 
     pub fn compute_is_persistent(block_utime: u32, prev_utime: u32) -> bool {
-        block_utime >> 17 != prev_utime >> 17
+        block_utime >> 10 != prev_utime >> 10
```
0xdeafbeef: Why these numbers?

SmaGMan: @drmick can you reply?

Rexagon (approved):

PR #559, opened 2025-01-22
Feat/msgs proc logic v3 fix buffer

drmick:


collator/src/collator/messages_reader/mod.rs
```diff
@@ -320,22 +322,27 @@ impl MessagesReader {
...
                     }
                 }
             }
+        }
 
-            let par_reader_state = par_reader.finalize(current_next_lt)?;
-            internals_reader_state
-                .partitions
-                .insert(par_id, par_reader_state);
+        let mut min_internals_processed_to = ProcessedTo::default();
```
SmaGMan, line 337: Create a separate function `MessageReader.get_min_internals_processed_to_by_shards`

collator/src/collator/messages_reader/mod.rs
```diff
@@ -361,14 +361,39 @@ impl MessagesReader {
...
+        let moved_accs_to_part_1 = Self::reset_partition_router_by_stats(
             &self.msgs_exec_params,
             &mut queue_diff_with_msgs.partition_router,
             aggregated_stats,
             self.for_shard_id,
             diffs,
         )?;
 
+        // remove moved accounts from partition 0 buffer
+        for (par_id, par_reader) in self.internals_partition_readers.iter_mut() {
```
SmaGMan, line 373: You can use `.get_mut(&0).unwrap()`

collator/src/collator/messages_reader/mod.rs
```diff
@@ -361,14 +361,39 @@ impl MessagesReader {
 
         // reset queue diff partition router
         // according to actual aggregated stats
-        Self::reset_partition_router_by_stats(
+        let moved_accs_to_part_1 = Self::reset_partition_router_by_stats(
```
SmaGMan, line 364: Let's rename to `moved_from_par_0_accounts`. Because further we may move not to only partition 1.

collator/src/collator/messages_reader/mod.rs
```diff
@@ -405,7 +431,8 @@ impl MessagesReader {
                         "move address {} to partition 1 because it has {} messages",
                         dest_int_address, msgs_count,
                     );
-                    partition_router.insert(RouterDirection::Dest, dest_int_address, 1)?;
+                    partition_router.insert(RouterDirection::Dest, dest_int_address.clone(), 1)?;
+                    moved_to_part_1.insert(dest_int_address.get_address());
```
SmaGMan, line 435: Better to save address in hasmap (`.get_address()`) first and then put `dest_int_address` to router to avoid cloning.

collator/src/collator/messages_reader/mod.rs
```diff
@@ -357,14 +359,43 @@ impl MessagesReader {
...
+        let moved_from_par_0_accounts = Self::reset_partition_router_by_stats(
             &self.msgs_exec_params,
             &mut queue_diff_with_msgs.partition_router,
             aggregated_stats,
             self.for_shard_id,
             diffs,
         )?;
 
+        // remove moved accounts from partition 0 buffer
+        let par_reader = self.internals_partition_readers.get_mut(&0).unwrap();
```
Rexagon, line 371: `if let Some(..) = self.internals_partition_readers.get_mut(&0)` just in case it is absent

Rexagon (changes requested):

collator/src/collator/messages_reader/mod.rs
SmaGMan: Partition 0 reader should exist anyway. Otherwise collator cannot work.

PR #565, opened 2025-01-24
feat(elect): specify msg fee in config

SmaGMan:


cli/src/cmd/elect/mod.rs
```diff
@@ -388,6 +397,7 @@ impl CmdOnce {
 
         let stake = self.stake.or(simple.stake).context("no stake specified")?;
         let stake_factor = self.stake_factor.or(simple.stake_factor);
+        let msg_fee = self.msg_fee.unwrap_or_else(|| simple.msg_fee());
```
0xdeafbeef, line 400: Maybe we need to add validation that `msg_fee` is not 0?

0xdeafbeef (approved):

Rexagon (approved):

PR #573, opened 2025-01-27
Fix moving block to archive

pashinov:


Rexagon (approved):

0xdeafbeef (approved):

Rexagon:
Don't merge yet, we need to check what happens if `move_into_archive` is called twice

PR #574, opened 2025-01-29
feat(storage): waiting for all split block tasks to be completed

pashinov:


storage/src/store/block/mod.rs
```diff
@@ -46,6 +46,7 @@ pub struct BlockStorage {
     block_subscriptions: SlotSubscriptions<BlockId, BlockStuff>,
     store_block_data: tokio::sync::RwLock<()>,
     prev_archive_commit: tokio::sync::Mutex<Option<CommitArchiveTask>>,
+    split_block_tasks: tokio::sync::Mutex<Vec<SplitBlockTask>>,
```
0xdeafbeef, line 49: You can use JoinSet instead of vec

PR #576, opened 2025-01-29
chore: GC metrics

drmick:


0xdeafbeef (approved):

core/src/block_strider/subscriber/gc_subscriber.rs
```diff
@@ -323,6 +325,9 @@ impl GcSubscriber {
                 }
             };
 
+            metrics::gauge!("tycho_core_last_mc_block_seqno_blocks_storage")
```
SmaGMan, line 328: Let's rename into `tycho_core_mc_blocks_gc_lag`

scripts/gen-dashboard.py
```diff
@@ -867,7 +867,13 @@ def storage() -> RowPanel:
                 )
             ],
             unit="Blocks",
```
SmaGMan, line 869: Let's set `units="States"`

scripts/gen-dashboard.py
```diff
@@ -867,7 +867,13 @@ def storage() -> RowPanel:
...
-            title="GC lag",
+            title="States GC lag",
+        ),
+        create_gauge_panel(
+            "tycho_core_last_mc_block_seqno_blocks_storage", "Blocks GC lag"
+        ),
```
SmaGMan, line 874: Let's add `units="Blocks"`

SmaGMan (approved):

PR #577, opened 2025-01-30
Blockchain service message limits

MrWad3r:


core/src/blockchain_rpc/service.rs
```diff
@@ -324,12 +325,34 @@ impl<B: BroadcastListener> Service<ServiceRequest> for BlockchainRpcService<B> {
...
+                                tracing::debug!("failed to validate external message: {e:?}");
+                            }
+                        }
+                    })
+                } else {
+                    BoxFutureOrNoop::future(async move {
+                        inner
+                            .broadcast_listener
+                            .handle_message(req.metadata, req.body)
+                            .await;
```
0xdeafbeef, line 353: Why don't you validate small messages?

MrWad3r: There is no need

0xdeafbeef: > There is no need

So I can send any garbage if its len < `BOUNDARY_BOC_SIZE` and you would gratefully broadcast it?

0xdeafbeef (approved):
lgtm

Rexagon (approved):

PR #578, opened 2025-01-30
Fix/collator large ranges

drmick:


collator/src/collator/messages_reader/internals_reader.rs
```diff
@@ -734,19 +781,20 @@ impl InternalsPartitionReader {
...
+                        reader.seqno,
                     ))
                 })
                 .unwrap_or_default();
-            let mut range_reader = self.create_next_internals_range_reader(
-                last_range_reader_shards_and_offset_opt,
-                self.block_seqno,
-            )?;
+            // we should look thru the whole range to check for pending messages
+            // so we do not pass `range_max_blocks` to force use the prev block end lt
```
SmaGMan, line 795: Now we do not pass `range_max_messages`. Please fix the comment

collator/tests/internal_queue.rs
```diff
@@ -174,7 +174,7 @@ fn test_statistics_check_statistics(
     assert_eq!(*addr_1_stat, 10000);
     assert_eq!(*addr_2_stat, 5000);
 
-    // check second diff
+    // check second diff, we have 0.,35000 lt in low partition
```
SmaGMan, line 177: "we have 0.,35000 lt in low partition"
what does it mean?

cli/src/cmd/tools/gen_zerostate.rs
```diff
@@ -701,6 +701,7 @@ fn make_default_params() -> Result<BlockchainConfigParams> {
             par_0_int_msgs_count_limit: 50_000,
             par_0_ext_msgs_count_limit: 5_000,
             group_slots_fractions,
+            range_messages_limit: 100_000,
```
SmaGMan, line 704: Let's set 10k as default.

collator/tests/internal_queue.rs
drmick: changed. it's comment on the test

SmaGMan (approved):

SmaGMan (approved):

PR #585, opened 2025-02-03
fix(collator): set gen_software to block from blockchain config

SmaGMan:
`CollatorConfig.supported_block_version` and `CollatorConfig.supported_capabilities` are now set up only in code.

We need to remove `collator.supported_block_version` and `collator.supported_capabilities` from the node's `config.json`.

### DIFF
--- a/collator/src/collator/do_collate/finalize.rs
+++ b/collator/src/collator/do_collate/finalize.rs
@@ -396,9 +396,12 @@ impl Phase<FinalizeState> {
         // info.want_split = false;
         // info.want_merge = false;
 
-        let capabilities = self.state.mc_data.config.get_global_version()?.capabilities;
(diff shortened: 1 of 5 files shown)


cli/src/node/config.rs
```diff
@@ -283,6 +283,7 @@ pub struct NodeConfig {
 
     pub archive_block_provider: ArchiveBlockProviderConfig,
 
+    #[serde(default)]
     pub collator: CollatorConfig,
```
0xdeafbeef, line 287: Won't it be inherited from the
```
#[serde(default)]
pub struct NodeConfig
```

0xdeafbeef (approved):

cli/src/node/config.rs
SmaGMan: Removed.

Rexagon (approved):

PR #591, opened 2025-02-04
cancel active collator sync when newer block received

SmaGMan:


collator/src/manager/blocks_cache.rs
```diff
@@ -520,16 +535,18 @@ impl BlocksCache {
...
+            // clean previous last collated blocks ids
+            guard
+                .data
+                .remove_prev_from_last_collated_block_ids(&block_id.seqno);
+
             occupied_entry
         };
 
-        self.extract_mc_block_subgraph(mc_block_entry)
+        Ok(self.extract_mc_block_subgraph(mc_block_entry))
```
drmick, line 546: Result in this func doesn't look necessary

drmick (approved):

PR #594, opened 2025-02-10
feature(collator): improve blocks sync speed

drmick:


collator/src/manager/mod.rs
```diff
@@ -469,23 +471,36 @@ where
                 candidate_stuff, ..
             } => (
                 &candidate_stuff.candidate.queue_diff_aug.data,
-                candidate_stuff
+                &candidate_stuff
                     .candidate
                     .block
                     .data
                     .load_extra()?
```
SmaGMan, line 478: It will be good to avoid pre-loading extra because diff could be skipped.

collator/src/manager/mod.rs
```diff
@@ -1253,6 +1277,27 @@ where
                     continue;
                 }
 
+                let top_block_shard = first_applied_top_blocks
+                    .get(&prev_block_id.shard)
+                    .unwrap_or(&0);
+
+                if prev_block_id.seqno < *top_block_shard {
```
SmaGMan, line 1284: Imaging. Node was stopped on the block 100. Then we restarted it on block 300. It loaded 200 blocks from archives and during this was not trying to sync. So your queue state is actual on the block 100. Then sync is running to block 301. And due to this logic node will skip all diffs below 301.

SmaGMan: Not sure if we need to compare with "top shard blocks" of the "first applied mc block" but we can take them easier from `BlockCacheEntry.top_shard_blocks_info`.

collator/src/manager/blocks_cache.rs
```diff
@@ -73,6 +71,31 @@ impl BlocksCache {
         Ok(result)
     }
 
+    pub fn get_top_shard_blocks(
```
SmaGMan, line 74: You can get "top shard blocks" simply from `BlockCacheEntry.top_shard_blocks_info` of master block.

collator/src/manager/mod.rs
```diff
@@ -1137,11 +1153,14 @@ where
 
     async fn sync_to_applied_mc_block_if_exist(
         &self,
-        _last_collated_block_id: Option<&BlockId>, // TODO: use to skip queue recovery
+        last_collated_block_id: Option<&BlockId>, // TODO: use to skip queue recovery
```
SmaGMan, line 1156: Pls, remove "TODO" comment now.

collator/src/manager/mod.rs
```diff
@@ -1240,6 +1260,33 @@ where
...
+                    for item in state.shards()?.iter() {
+                        let (shard_id, shard_descr) = item?;
+                        lower_border.insert(shard_id, shard_descr.get_block_id(shard_id).seqno);
+                    }
+                }
+                Some(top_shard_blocks) => {
+                    lower_border.extend(top_shard_blocks);
+                }
+            }
+        }
```
SmaGMan, line 1288: Pls, extract this to a separate method. And fix typo in the comment. Rename `lower_border` to `top_collated_blocks`.

collator/src/manager/mod.rs
```diff
@@ -1352,13 +1405,24 @@ where
 
             // apply queue diffs
```
SmaGMan, line 1406: Here you should compare with top collated blocks as well. It could be a situation when we collated blocks 1, 2, 3, 4. Then received 1, 2 (collated are correct). Then received 3, 4 (collated are not correct). On block 4 received range will be (1..4), but diffs 1, 2 are already applied.

SmaGMan:
will try to create again due to errors in checks

PR #599, opened 2025-02-12
Internal queue state improving & partitions

drmick:
- Store/Load last committed masterchain block id
- Fix diff tail trimming
- Move all in-memory internal queue data (diffs/tail) to RocksDB
- Remove uncommitted state. Use one column famity for committed and uncommitted data divided by commit pointer
- Introduced cumulative queue stats. Use cumulative stats during collation for isolated partitions.
- Implemented test of queue refill on sync to master block from blockchain
- Parititons routing
- Created new tables for statistics(`int_msg_statistics`) and internal messages (`shard_int_messages`). Old tables should be deleted by separated patch. Also should be deleted unused uncommitted state tables.
- Inizializing an internal queue from persistent state also initializes data for new tables.
- `CollatorConfig.fast_sync = true` by default

**IMPORTANT**: Node must be patched with an empty internal queue and no load


collator/src/manager/mod.rs
```diff
@@ -1040,7 +1050,11 @@ where
     /// Finish active sync if it is not finished yet
     /// and enqueue received block
     #[tracing::instrument(skip_all, fields(block_id = %state.block_id().as_short_id()))]
-    pub async fn enqueue_handle_block_from_bc(&self, state: ShardStateStuff) -> Result<()> {
+    pub async fn enqueue_handle_block_from_bc(
+        &self,
+        state: ShardStateStuff,
```
SmaGMan, line 1055: Let's use some context struct to pass state and processed_upto into methods and thru the channel instead of adding additional arg in each method.

collator/src/collator/tests/messages_reader_tests.rs
```diff
@@ -1484,11 +1483,9 @@ impl std::fmt::Debug for TestInternalMessageType {
 async fn create_test_queue_adapter<V: InternalMessageValue>(
 ) -> Result<(Arc<dyn MessageQueueAdapter<V>>, tempfile::TempDir)> {
     let (storage, tmp_dir) = Storage::new_temp().await?;
-    let uncommitted_state_factory = UncommittedStateImplFactory::new(storage.clone());
-    let committed_state_factory = CommittedStateImplFactory::new(storage.clone());
+    let committed_state_factory = QueueStateImplFactory::new(storage.clone());
```
SmaGMan, line 1486: Should rename `committed_state_factory`

SmaGMan (changes requested):

collator/src/internal_queue/queue.rs
```diff
@@ -232,6 +239,7 @@ where
             self,
             &block_id_short.shard,
             block_id_short.seqno,
+            DiffZone::Uncommitted,
```
SmaGMan, line 242: Seems like safer to look in both zones.

collator/src/internal_queue/queue.rs
```diff
@@ -379,33 +383,30 @@ where
         Ok(statistics)
     }
 
-    fn get_diff(&self, shard_ident: &ShardIdent, seqno: u32) -> Result<Option<DiffInfo>> {
-        self.state.get_diff_info(shard_ident, seqno)
+    fn get_diff(
```
SmaGMan, line 386: Let's rename to `get_diff_info`

collator/src/manager/mod.rs
```diff
@@ -1627,17 +1643,25 @@ where
...
             }
-        }
 
-        // apply required previous queue diffs
-        while let Some((diff, diff_hash, block_id, min_message, max_message)) =
-            prev_queue_diffs.pop()
-        {
-            let statistics =
-                DiffStatistics::from_diff(&diff, block_id.shard, min_message, max_message);
+            let mut is_first_diff = true;
```
SmaGMan, line 1647: We should skip not for first applied but for first required. If we have such diffs in queue: 10, 11, 12, 13. And we sync to 17 which requires diffs 12-16. Then we will not add 12 and 13 to `prev_queue_diffs`. The first one in `prev_queue_diffs` will be 14. And we should perform check for it. We should not perform check only for 12 if we apply it.

SmaGMan: We need to add a unit test to cover the queue init on sync logic. Currently `test_version()` and `test_queue()` do not cover this case.

collator/src/manager/mod.rs
```diff
@@ -1627,17 +1643,25 @@ where
...
+            // apply required previous queue diffs for each shard
+            while let Some((diff, diff_hash, block_id, min_message, max_message)) =
+                prev_queue_diffs.pop()
+            {
+                let statistics =
+                    DiffStatistics::from_diff(&diff, block_id.shard, min_message, max_message);
+
+                self.mq_adapter.apply_diff(
```
SmaGMan, line 1655: We need to add a unit test to cover the queue init on sync logic. Currently `test_version()` and `test_queue()` do not cover this case.

collator/src/manager/mod.rs
```diff
@@ -2744,6 +2796,18 @@ where
             }) => {
                 extract_elapsed = histogram_extract.finish();
 
+                let mut partitions = FastHashSet::default();
```
SmaGMan, line 2799: We should use `subgraph.get_partitions()` here as well.

collator/src/manager/mod.rs
```diff
@@ -756,6 +765,20 @@ where
...
+            collation_result
+                .candidate
+                .processed_upto
+                .partitions
+                .keys()
+                .copied(),
+        );
+
+        if let Some(mc_data) = &collation_result.mc_data {
+            partitions.extend(mc_data.processed_upto.partitions.keys().copied());
```
SmaGMan, line 779: For master block it will be equal to `collation_result.candidate.processed_upto.partitions`. So this extend is redundant.

collator/src/manager/mod.rs
```diff
@@ -756,6 +765,20 @@ where
             collation_result.mc_data.is_some(),
         );
 
+        let mut partitions = FastHashSet::default();
+        partitions.extend(
+            collation_result
+                .candidate
+                .processed_upto
+                .partitions
```
SmaGMan, line 773: You created method `get_partitions()`. You can use it here.

collator/src/manager/mod.rs
```diff
@@ -1136,7 +1163,12 @@ where
     /// 4. Notify mempool about new master block
     /// 5. Sync to received block if it is far ahead last collated and last synced_to
     #[tracing::instrument(skip_all, fields(block_id = %state.block_id().as_short_id()))]
-    async fn handle_block_from_bc(&self, state: ShardStateStuff, skip_sync: bool) -> Result<()> {
+    async fn handle_block_from_bc(
+        &self,
+        state: ShardStateStuff,
```
SmaGMan, line 1168: Let's use `HandledBlockFromBcCtx` here as well.

collator/src/manager/mod.rs
```diff
@@ -1145,13 +1177,22 @@ where
 
         let _histogram = HistogramGuard::begin("tycho_collator_handle_block_from_bc_time");
 
+        let mut partitions = processed_upto.get_partitions();
+        for partition in state.state().processed_upto.load()?.partitions.keys() {
```
SmaGMan, line 1181: It is the same as `processed_upto`.

collator/src/manager/mod.rs
```diff
@@ -1208,8 +1221,6 @@ where
                 }
             }
 
-            self.mq_adapter.clear_uncommitted_state()?;
```
SmaGMan, line 1187: Why this was removed?

collator/src/manager/mod.rs
```diff
@@ -1521,6 +1533,9 @@ where
             FastHashMap::default()
         };
 
+        // clear uncommitted state before applying diffs
+        self.mq_adapter.clear_uncommitted_state()?;
```
SmaGMan, line 1537: Looks like we may have a bug here. Imaging that we have committed diff 100 and applied 101, 102, 103. Then we read diff 103 into `queue_diffs_applied_to_top_blocks`. Then we remove uncommitted diffs 101-103. Then we read diff 101, see that it below 103 and skip them. So we will have a hole in the diffs sequence.

collator/src/collator/do_collate/finalize.rs
```diff
@@ -175,13 +216,16 @@ impl Phase<FinalizeState> {
...
                 );
 
-                mq_adapter.apply_diff(
-                    queue_diff_with_msgs,
-                    block_id_short,
-                    &queue_diff_hash,
-                    statistics,
-                    max_message,
-                )?;
+                // TODO should check sequence only if previous diff is required or check only uncommitted zone
```
SmaGMan, line 219: Needs to remove this TODO. The `DiffZone::Uncommitted` is preferred here.

collator/src/collator/messages_reader/mod.rs
```diff
@@ -433,21 +464,29 @@ impl<V: InternalMessageValue> MessagesReader<V> {
...
+            .cloned()
+            .unwrap_or_default();
 
         // reset queue diff partition router
         // according to actual aggregated stats
         let moved_from_par_0_accounts = Self::reset_partition_router_by_stats(
             &self.msgs_exec_params,
             &mut queue_diff_with_msgs.partition_router,
-            aggregated_stats,
+            self.internal_queue_statistics.get_aggregated_result(),
```
SmaGMan, line 485: We should append stats from new messages (aggregated diff stats) to aggregated cumulative stats to take them into account when detecting target partition for diff. Without this we consider new messages from block A only on the next block A+1.

[related issue](https://github.com/broxus/tycho/issues/683)

collator/src/collator/mod.rs
```diff
@@ -1300,11 +1323,16 @@ impl CollatorStdImpl {
                     .iter()
                     .map(|(k, v)| (*k, v.end_lt))
                     .collect(),
+                all_shards_processed_to_by_partitions,
```
SmaGMan, line 1326: We do not need to load cumulative stats here because it is not used to check if any message exists in iterators. Additionally if we run this code for master block we do not know the actual top shard blocks. So if we load and store cumulative stats it will be incorrect for further collation.

drmick: Probably we can use `all_shards_processed_to_by_partitions: Default::default()`

SmaGMan (approved):

collator/src/collator/do_collate/finalize.rs
```diff
@@ -411,9 +411,18 @@ impl Phase<FinalizeState> {
             (None, Some(self.state.mc_data.make_block_ref()))
         };
 
+        // HACK: make every 3-d block incorrect on debug if env variable defined
+        let version =
+            if cfg!(debug_assertions) && self.state.collation_data.block_id_short.seqno % 3 == 0 {
```
0xdeafbeef, line 416: release_check profile sets it. Better to use tycho_unstable  and run tests with
`RUSTFLAGS="--cfg tycho_unstable` cargo ...`

collator/src/collator/do_collate/finalize.rs
```diff
@@ -169,6 +169,8 @@ impl Phase<FinalizeState> {
             }
         };
 
+        //// !!! обновить ридер стейт
```
0xdeafbeef: :thinking:

collator/src/collator/do_collate/finalize.rs
```diff
@@ -169,8 +169,6 @@ impl Phase<FinalizeState> {
             }
         };
 
-        //// !!! обновить ридер стейт
```
0xdeafbeef: nvm, here it's deleted

collator/src/collator/do_collate/finalize.rs
SmaGMan: fixed

0xdeafbeef (approved):

Rexagon (approved):

PR #600, opened 2025-02-12
fix(collator): remove all collated block ids from mismatched one

SmaGMan:


SmaGMan:
Included in https://github.com/broxus/tycho/pull/606

PR #644, opened 2025-03-31
ref(collator): move internal border logic to storage

drmick:


collator/src/queue_adapter.rs
```diff
@@ -143,9 +154,18 @@ impl<V: InternalMessageValue> MessageQueueAdapter<V> for MessageQueueAdapterStdI
     ) -> Result<QueueStatistics> {
         let start_time = std::time::Instant::now();
 
-        let stats = self.queue.load_diff_statistics(partitions, ranges)?;
+        let mut result = AccountStatistics::default();
 
-        let stats = QueueStatistics::with_statistics(stats);
+        for range in ranges.to_vec().iter_mut() {
+            for partition in partitions {
+                range.from = range.from.next_value();
```
SmaGMan, line 161: Here we will take next many times inside the cycle by partitions.

PR #649, opened 2025-04-01
wip: use all shards processed to info detailed by partitions

SmaGMan:


collator/src/collator/do_collate/mod.rs
```diff
@@ -336,6 +336,7 @@ impl CollatorStdImpl {
                 queue_diff_messages_count,
                 has_unprocessed_messages,
                 reader_state,
+                processed_upto,
```
drmick, line 339:
```suggestion
                mut processed_upto,
```

collator/src/collator/do_collate/mod.rs
```diff
@@ -344,7 +345,7 @@ impl CollatorStdImpl {
 
         let finalize_block_timer = std::time::Instant::now();
 
-        let mut processed_upto = reader_state.get_updated_processed_upto();
+        let mut processed_upto = processed_upto;
```
drmick, line 348:
```suggestion
```

collator/src/types/processed_upto.rs
```diff
@@ -368,18 +376,64 @@ impl ProcessedUptoInfoExtension for ProcessedUptoInfoStuff {
...
+            for (&shard_id, &to_key) in partition_processed_to {
+                result
+                    .entry(shard_id)
+                    .and_modify(|min| *min = std::cmp::min(*min, to_key))
+                    .or_insert(to_key);
+            }
+        }
+    }
+
+    result
```
drmick, line 438: Is it okay that we have here infromation only from masterchain and current shard. Should we add `processed to` from other mc_data shards?

collator/src/types/processed_upto.rs
```diff
@@ -368,18 +376,64 @@ impl ProcessedUptoInfoExtension for ProcessedUptoInfoStuff {
...
+    all_shards_processed_to_by_partitions: &FastHashMap<
+        ShardIdent,
+        (bool, ProcessedToByPartitions),
+    >,
+) -> ProcessedTo {
+    let mut result = ProcessedTo::default();
+
+    for (_, shard_processed_to_by_partitions) in all_shards_processed_to_by_partitions
+        .values()
+        .filter(|(updated, _)| *updated)
```
drmick, line 426: The previous function version didn't have an `update` check. This check is ok for `Self::restore_queue`?

collator/src/types/processed_upto.rs
SmaGMan: This method designed to be used with a map that contains all shards data including master and current shard.

collator/src/types/processed_upto.rs
SmaGMan: Yes. Because previously we created the map with only updated shards with the previous version of method `get_all_shards_processed_to_by_partitions_for_mc_block()` and then passed it to `find_min_processed_to_by_shards()`. Now method `get_all_shards_processed_to_by_partitions_for_mc_block()` returns all shards with an `updated` flag.

PR #657, opened 2025-04-04
use cumulative stats on messages collection from buffers

SmaGMan:


collator/src/collator/messages_reader/internals_reader.rs
```diff
@@ -664,7 +675,7 @@ impl<V: InternalMessageValue> InternalsPartitionReader<V> {
 
                             metrics.read_existing_msgs_count += 1;
 
-                            // update remaining messages statistics in iterator
+                            // update remaning messages statistics in range reader
```
drmick, line 678: `remaining` is correct

collator/src/collator/types.rs
```diff
@@ -1257,14 +1265,19 @@ impl CumulativeStatistics {
         diff_max_message: DiffMaxMessage,
         diff_partition_stats: AccountStatistics,
     ) {
+        // append to cumulative stats
+        if let Some(partition_stats) = self.result.get_mut(&partition) {
```
drmick, line 1269: The `result` is filled in the `recalculate` method, but `dirty` is always `false` and `recalculate` will never execute

collator/src/collator/types.rs
```diff
@@ -1257,14 +1265,19 @@ impl CumulativeStatistics {
...
+
+        // update diffs stats collection
         self.shards_stats_by_partitions
             .entry(diff_shard)
             .or_default()
             .entry(partition)
             .or_default()
             .insert(diff_max_message, diff_partition_stats);
-
-        self.dirty = true;
```
drmick, line 1267: Why did you remove this?

collator/src/collator/types.rs
```diff
@@ -1296,17 +1309,25 @@ impl CumulativeStatistics {
                 if let Some(partition_processed_to) =
                     shard_processed_to_by_partitions.get(partition)
                 {
+                    let cumulative_stats = self.result.get_mut(partition).unwrap();
```
drmick, line 1312: The `result` is filled in the `recalculate` method, but `dirty` is always `false` and `recalculate` will never execute

collator/src/collator/types.rs
SmaGMan: Because I am appending stats in result with stats from the diff. So we do not need to recalculate result.

collator/src/collator/messages_reader/internals_reader.rs
SmaGMan: fixed

collator/src/collator/types.rs
SmaGMan: fixed

collator/src/collator/types.rs
SmaGMan: fixed

PR #673, opened 2025-04-16
refactor(consensus): simplify query wrapper

Mododo:
Exposes `Point` deserialization and hash check, so they can be delegated to rayon. Also raw point bytes can be preserved until inserted into db.


MrWad3r (approved):
lgtm. Some tests would be appreciated to be sure everything works correctly

Rexagon (approved):

PR #684, opened 2025-04-17
chore(collator): move calc `shards_processed_to_by_partitions`

drmick:


collator/src/collator/do_collate/mod.rs
```diff
@@ -725,6 +730,13 @@ impl CollatorStdImpl {
                 collation_data_builder.store_shard_fees(shard_id, proof_funds)?;
             }
 
+            collation_data_builder
```
SmaGMan, line 733: This way we will loose info from not updated shards, because `top_shard_blocks_info` will not contain data for shard if newer block was not collated between master blocks.

SmaGMan (changes requested):
Fix required (see comments)

SmaGMan (approved):
Squash and merge

PR #693, opened 2025-04-21
fix(collator): add new messages to aggregate stats

drmick:


collator/src/collator/messages_reader/mod.rs
```diff
@@ -480,12 +480,26 @@ impl<V: InternalMessageValue> MessagesReader<V> {
             .cloned()
             .unwrap_or_default();
 
+        let internals_processed_to =
+            Self::collect_internals_processed_to(&self.internals_partition_readers);
```
SmaGMan, line 484: Let's make it the owned function of MessagesReader.

collator/src/collator/messages_reader/mod.rs
```diff
@@ -480,12 +480,26 @@ impl<V: InternalMessageValue> MessagesReader<V> {
...
+
+        // reduce stats of processed diffs
+        self.internal_queue_statistics
+            .handle_processed_to_update(self.for_shard_id, internals_processed_to);
+
+        let mut aggregated_result = self.internal_queue_statistics.get_aggregated_result();
```
SmaGMan, line 490: Let's rename to `aggregated_stats`

SmaGMan (approved):

PR #708, opened 2025-04-29
limit accumulated wu

SmaGMan:
**TESTS**
**Unit Tests**
No changes
**Network Tests**
No special coverage
**Manual Tests**
Tested by 20k transfers (run-everwallets20). Time diff jumps removed. See screenshots below.

**PERFORMANCE IMPACT**
Expected impact
Collator should process messages normally when heavy load finished.

**NODE CONFIGURATION CHANGES**
None
**BLOCKCHAIN CONFIGURATION CHANGES**
None

**COMPATIBILITY**
Collator. Work Units Calculation - fully compatible
We limit max accumulated wu to cover only 210 rounds. Under heavy load if some node is late (e.g. collating mb 100 when others collating mb 110) then it will produce mismatched mb 100 and then sync to the last one. All nodes in consensus will produce identical next blocks after restart.

**SPECIAL DEPLOYMENT ACTIONS**
Required
Better to perform update when no heavy load, externals queue is not very big, and dag consensus lag is below 100.

### DIFF
--- a/collator/src/collator/do_collate/finalize.rs
+++ b/collator/src/collator/do_collate/finalize.rs
@@ -486,20 +486,38 @@ impl Phase<FinalizeState> {
             );
 
             // compute total wu used from last anchor
-            let new_wu_used_from_last_anchor = wu_used_from_last_anchor
+            let mut new_wu_used_from_last_anchor = wu_used_from_last_anchor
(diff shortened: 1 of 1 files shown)


SmaGMan:
Before
image

After
image

Time diff becomes normal when load finished.

Rexagon (approved):

0xdeafbeef (approved):

PR #725, opened 2025-05-05
feat(collator): add batch modification of shard accounts dict

### LINKED ISSUE #716: Somehow consume accounts inside an iterator when building `account_blocks`
https://github.com/broxus/tycho/blob/dd9f5c82c9359f84ce8df75ad6a9e260684ffbc8/collator/src/collator/do_collate/finalize.rs#L1104

### LINKED ISSUE #724: Optimize building shard accounts during block finalization
For now, we're rebuilding the account dictionary up to the root on every changed/removed account. We can speed this up significantly by rebuilding the array directly with the new elements already included/excluded. The complexity should become linear, and perf could increase by up to 2x. 

This will allow us to use more complex structs as extra values (unblocks #720).

Rexagon:
Closes #724 
Closes #716

And fixes extra currencies support (there were situations where wrong extra value offset could lead to incorrect values merge).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -956,7 +956,7 @@ dependencies = [
 [[package]]
 name = "everscale-types"
 version = "0.1.2"
-source = "git+https://github.com/broxus/everscale-types.git?rev=82c343b2bd17ceb51db0575b6b8ee821ccda037e#82c343b2bd17ceb51db0575b6b8ee821ccda037e"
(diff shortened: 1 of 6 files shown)


Rexagon:
Metric `only build account blocks`

Before:
image

After:
image

0xdeafbeef (approved):

SmaGMan (approved):

PR #748, opened 2025-05-16
move blobs from rocksdb-blobdb to cas based storage

0xdeafbeef:
Rocsdkb paired with blobdb results in 5-25 write amplification, which easily saturates write bandwidth and does not offer any improvements compared to the in-house cas based blob storage. Also nvme will die. BADBADNOTGOOD.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -428,7 +428,7 @@ dependencies = [
 [[package]]
 name = "cassadilia"
 version = "0.1.0"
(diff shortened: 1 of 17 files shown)


0xdeafbeef:
left one is current branch; right one is dd5685d8
image

image


- [x] Sync of new version against master working after node stop + start.
- [x] Sync of new version against new version  is working after node stop + start.
```
du -sh db
95G     db
```
- [x] Sync of master version against new is working after node stop + start.
```
du -sh db                                                          
167G    db
```

https://grafana.broxus.com/goto/b_uSAzfNR?orgId=1

PR #754, opened 2025-05-21
dump state tool and test collate block from dump

SmaGMan:


SmaGMan:
Will be implemented in #891

PR #768, opened 2025-05-27
feature(collator): append cumulative statistics after new master block

drmick:
NO
NO
Not Required
### PERFORMANCE IMPACT
Now statistics is reusing from previous collation and full reload statistics doesn't necessary
Metrtics `tycho_internal_queue_separated_statistics_load_time` should looks better

1-255 10k + transfers 20k:
image

---

#### Unit Tests

`cargo test --lib collator::messages_reader::tests::test_refill_messages`

#### Network Tests

No covarage

#### Manual Tests
[1-255 10k + transfers 20k ](https://grafana.broxus.com/d/cdlaji62a1b0gb/tycho-node-metrics-test?orgId=1&from=2025-06-05T11:22:27.237Z&to=2025-06-05T11:58:34.522Z&timezone=browser&var-source=ced0ihciqrn5sb&var-instance=tycho-devnet2-validator1&var-workchain=$__all&var-partition=$__all&var-kind=$__all&var-method=$__all&var-peer_id=$__all&var-remote_addr=$__all&refresh=30s)

### DIFF
--- a/collator/src/collator/do_collate/mod.rs
+++ b/collator/src/collator/do_collate/mod.rs
@@ -34,7 +34,7 @@ use crate::queue_adapter::MessageQueueAdapter;
 use crate::tracing_targets;
 use crate::types::{
     BlockCollationResult, BlockIdExt, CollationSessionInfo, CollatorConfig,
-    DisplayBlockIdsIntoIter, DisplayBlockIdsIter, McData, McDataStuff, ProcessedToByPartitions,
+    DisplayBlockIdsIntoIter, DisplayBlockIdsIter, McData, ProcessedToByPartitions,
     ShardDescriptionShort, ShardDescriptionShortExt, TopBlockDescription, TopShardBlockInfo,
 };
 
@@ -71,7 +71,7 @@ impl CollatorStdImpl {
 
         let WorkingState {
             next_block_id_short,
-            mc_data_stuff,
+            mc_data,
             collation_config,
             wu_used_from_last_anchor,
             prev_shard_data,
@@ -80,14 +80,14 @@ impl CollatorStdImpl {
             ..
         } = *working_state;
 
-        let mc_block_id = mc_data_stuff.current.block_id;
+        let mc_block_id = mc_data.block_id;
         let prev_shard_data = prev_shard_data.unwrap();
         let usage_tree = usage_tree.unwrap();
         let tracker = prev_shard_data.ref_mc_state_handle().tracker().clone();
 
         tracing::info!(target: tracing_targets::COLLATOR,
             "Start collating block: mc_data_block_id={}, prev_block_ids={}, top_shard_blocks_ids: {:?}",
-            mc_data_stuff.current.block_id.as_short_id(),
+            mc_data.block_id.as_short_id(),
             DisplayBlockIdsIntoIter(prev_shard_data.blocks_ids()),
             top_shard_blocks_info.as_ref().map(|v| DisplayBlockIdsIter(
                 v.iter().map(|i| &i.block_id)
@@ -106,29 +106,25 @@ impl CollatorStdImpl {
 
         let is_first_block_after_prev_master = is_first_block_after_prev_master(
             prev_shard_data.blocks_ids()[0], // TODO: consider split/merge
-            &mc_data_stuff.current.shards,
+            &mc_data.shards,
         );
-        let part_stat_ranges =
-            if is_first_block_after_prev_master && mc_data_stuff.current.block_id.seqno > 0 {
-                if next_block_id_short.is_masterchain() {
-                    self.mc_compute_part_stat_ranges(
-                        &mc_data_stuff,
-                        top_shard_blocks_info.clone().unwrap(),
-                    )
+        let part_stat_ranges = if is_first_block_after_prev_master && mc_data.block_id.seqno > 0 {
+            if next_block_id_short.is_masterchain() {
+                self.mc_compute_part_stat_ranges(&mc_data, top_shard_blocks_info.clone().unwrap())
                     .await?
-                } else {
-                    self.compute_part_stat_ranges(&mc_data_stuff, &next_block_id_short)
-                        .await?
-                }
             } else {
-                None
-            };
+                self.compute_part_stat_ranges(&mc_data, &next_block_id_short)
+                    .await?
+            }
+        } else {
+            None
+        };
 
         let collation_data = self.create_collation_data(
             next_block_id_short,
             next_chain_time,
             created_by,
-            &mc_data_stuff.current,
+            &mc_data,
             &prev_shard_data,
             top_shard_blocks_info,
         )?;
@@ -139,7 +135,7 @@ impl CollatorStdImpl {
         let state = Box::new(ActualState {
             collation_config,
             collation_data,
-            mc_data_stuff,
+            mc_data,
             prev_shard_data,
             shard_id: self.shard_id,
             collation_is_cancelled: CancellationFlag::new(),
@@ -307,7 +303,7 @@ impl CollatorStdImpl {
     ) -> Result<CollationResult, CollatorError> {
         let shard_id = state.shard_id;
         let labels = [("workchain", shard_id.workchain().to_string())];
-        let mc_data_stuff = state.mc_data_stuff.clone();
+        let mc_data = state.mc_data.clone();
 
         let collation_is_cancelled = state.collation_is_cancelled.clone();
 
@@ -403,8 +399,7 @@ impl CollatorStdImpl {
             .get_min_internals_processed_to_by_shards()
             .get(&shard_id)
             .cloned();
-        let min_processed_to_from_mc_data = mc_data_stuff
-            .current
+        let min_processed_to_from_mc_data = mc_data
             .processed_upto
             .get_min_internals_processed_to_by_shards()
             .get(&shard_id)
@@ -957,8 +952,8 @@ impl CollatorStdImpl {
 
             let new_queue_diff_hash = *finalized.block_candidate.queue_diff_aug.diff_hash();
 
-            let collation_config = match &finalized.mc_data_stuff {
-                Some(mcd) => Arc::new(mcd.current.config.get_collation_config()?),
+            let collation_config = match &finalized.mc_data {
+                Some(mcd) => Arc::new(mcd.config.get_collation_config()?),
                 None => finalized.collation_config,
             };
 
@@ -967,16 +962,14 @@ impl CollatorStdImpl {
                 .on_block_candidate(BlockCollationResult {
                     collation_session_id: self.collation_session.id(),
                     candidate: finalized.block_candidate,
-                    prev_mc_block_id: finalized.old_mc_data_stuff.current.block_id,
-                    mc_data_stuff: finalized.mc_data_stuff.clone(),
+                    prev_mc_block_id: finalized.old_mc_data.block_id,
+                    mc_data: finalized.mc_data.clone(),
                     collation_config: collation_config.clone(),
                     force_next_mc_block,
                 })
                 .await?;
 
-            let new_mc_data_stuff = finalized
-                .mc_data_stuff
-                .unwrap_or(finalized.old_mc_data_stuff);
+            let new_mc_data = finalized.mc_data.unwrap_or(finalized.old_mc_data);
 
             // spawn update PrevData and working state
             self.prepare_working_state_update(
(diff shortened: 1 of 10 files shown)


collator/src/collator/do_collate/mod.rs
```diff
@@ -1189,6 +1206,204 @@ impl CollatorStdImpl {
...
+        mc_data: &McData,
+        block_id_short: &BlockIdShort,
+    ) -> Result<Option<Vec<QueueShardBoundedRange>>> {
+        let mc_block_id = mc_data.block_id;
+
+        let prev_mc_block_id = mc_data
+            .prev_mc_block_id
+            .context("Prev MC block must be present")?;
+
+        if prev_mc_block_id.seqno + 1 != mc_block_id.seqno {
```
SmaGMan, line 1224: Is it really possible? Maybe just add assert?

collator/src/collator/do_collate/mod.rs
```diff
@@ -1189,6 +1206,204 @@ impl CollatorStdImpl {
             "total collation timings"
         );
     }
+
+    /// Collect ranges for loading statistics if current block is first after master
+    /// Using previous masterchain block diff for range
+    /// and other shards diffs between previous masterchain block - 1 and previous masterchain block
+    async fn compute_part_stat_ranges(
```
SmaGMan, line 1213: Looks very similar to `mc_compute_part_stat_ranges()`. Can we merge them?

collator/src/collator/do_collate/mod.rs
```diff
@@ -1189,6 +1206,204 @@ impl CollatorStdImpl {
...
+        if prev_mc_block_id.seqno + 1 != mc_block_id.seqno {
+            tracing::error!(
+                target: tracing_targets::COLLATOR,
+                "Prev MC block ID has an incorrect sequence. Prev: {prev_mc_block_id:?}. \
+             Current: {mc_block_id:?}"
+            );
+            return Ok(None);
+        }
+
+        let prev_mc_state = self
```
SmaGMan, line 1233: Seems like it is better to store the previous McData value in the current McData? We can create
```
McDataStuff {
  current: Arc<McData>,
  prev: Option<Arc<McData>>
}
```
and pass it in most places where `Arc<McData>` is passed now.

So, we can create next McData in the `finalize()`
```
let mc_data_stuff = McDataStuff {
  current: Arc::new(McData { ... }),
  previous: Some(mc_data)
}
```

When we create mc_data on sync we can create it with `previous: None`. Because we have to load cumulative stats from the scratch after sync.

collator/src/collator/do_collate/mod.rs
```diff
@@ -1189,6 +1206,204 @@ impl CollatorStdImpl {
...
+                from: Bound::Excluded(first_diff_msg),
+                to: Bound::Included(last_diff_msg),
+            });
+        }
+
+        Ok(Some(ranges))
+    }
+
+    /// Helper function that retrieves `max_message` from either MQ or state diff
+    async fn get_max_message(&self, block_id: &BlockId) -> Result<Option<QueueKey>> {
```
SmaGMan, line 1394: Suggest to rename into `get_diff_max_message()`

collator/src/collator/messages_reader/internals_reader.rs
```diff
@@ -354,11 +354,11 @@ impl<V: InternalMessageValue> InternalsPartitionReader<V> {
                 fully_read = false;
             }
 
-            ranges.push(QueueShardRange {
+            ranges.push(QueueShardBoundedRange {
                 shard_ident: *shard_id,
                 // TODO: may be we should pass `from` here because we should load stats for the full range
```
SmaGMan, line 359: Is this todo actual here regarding the changes?

collator/src/collator/messages_reader/mod.rs
```diff
@@ -201,31 +203,61 @@ impl<V: InternalMessageValue> MessagesReader<V> {
...
-                inner_cumulative_statistics.load(
-                    mq_adapter.clone(),
-                    &cx.for_shard_id,
-                    &partitions,
-                    cx.prev_state_gen_lt,
-                    cx.mc_state_gen_lt,
-                    &cx.mc_top_shards_end_lts.iter().copied().collect(),
-                )?;
+
                 cumulative_stats_just_loaded = true;
```
SmaGMan, line 220: Is it correct to set `cumulative_stats_just_loaded = true` when cumulative stats is partially appended?

collator/src/collator/messages_reader/mod.rs
```diff
@@ -201,31 +203,61 @@ impl<V: InternalMessageValue> MessagesReader<V> {
...
-                inner_cumulative_statistics
+                match (previous_cumulative_statistics, cx.part_stat_ranges) {
+                    (Some(mut previous_cumulative_statistics), Some(part_stat_ranges)) => {
+                        // update all_shards_processed_to_by_partitions and truncate processed data
+                        previous_cumulative_statistics.update_processed_to_by_partitions(
+                            params.all_shards_processed_to_by_partitions.clone(),
+                        );
+
+                        // truncate data before range
+                        previous_cumulative_statistics.truncate_before(
```
SmaGMan, line 229: Does not this method duplicate what `update_processed_to_by_partitions()` do?

collator/src/collator/types.rs
```diff
@@ -1344,6 +1379,70 @@ impl CumulativeStatistics {
...
+
+            if changed {
+                self.handle_processed_to_update(dst_shard, processed_to.clone());
+            }
+        }
+
+        // TODO should remove full stat by shard if it doesn't exist in new_pt?
+
+        self.all_shards_processed_to_by_partitions = new_pt;
+        self.dirty = true;
```
SmaGMan, line 1443: Can we avoid recalculation?

collator/src/collator/types.rs
```diff
@@ -1402,6 +1501,9 @@ impl CumulativeStatistics {
                 }
             }
         }
+        if diff_partition_stats.is_empty() {
```
SmaGMan, line 1504: What it for? Let's add the comment describing the reason.

collator/src/collator/messages_reader/mod.rs
drmick: The enriched data should be equal to the data loaded from scratch. Therefore, for further processing, there should be no difference where the data was obtained from.

collator/src/collator/messages_reader/internals_reader.rs
drmick: `current_position` has already processed key, I'm not sure that we should include processed key to range for loading stats

collator/src/collator/messages_reader/mod.rs
drmick: Yes, its similar. truncate_before is removed

collator/src/collator/types.rs
```diff
@@ -1344,6 +1379,70 @@ impl CumulativeStatistics {
...
+            };
+
+            if changed {
+                self.handle_processed_to_update(dst_shard, processed_to.clone());
+            }
+        }
+
+        // TODO should remove full stat by shard if it doesn't exist in new_pt?
+
+        self.all_shards_processed_to_by_partitions = new_pt;
```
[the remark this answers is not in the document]
drmick: `handle_processed_to_update` only works with `new_pt` data, but `new_pt` may contain data that is not in `all_shards_processed_to_by_partitions`. It is better to set the new value directly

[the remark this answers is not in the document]
SmaGMan: If `all_shards_processed_to_by_partitions` does not contain shard from `new_pt` then `changed == true` that will cause adding new `ProcessedToByPartitions` into `all_shards_processed_to_by_partitions` inside `handle_processed_to_update`. Aren't this work like this?

collator/src/types.rs
```diff
@@ -439,10 +463,10 @@ impl BlockIdExt for BlockIdShort {
     }
 }
 
-pub trait ShardDescriptionExt {
+pub trait ShardDescriptionShortExt {
```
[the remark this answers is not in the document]
drmick: We have
`impl ShardDescriptionShortExt for ShardDescriptionShort {`

we can't implement `ShardDescriptionExt` for `ShardDescriptionShort`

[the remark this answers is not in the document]
SmaGMan: Got it. Let's keep as is.

collator/src/collator/types.rs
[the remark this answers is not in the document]
drmick: Refactored

collator/src/collator/types.rs
drmick: Refactored

collator/src/collator/types.rs
```diff
@@ -1279,6 +1280,9 @@ pub struct QueueStatisticsWithRemaning {
 }
 
 pub struct CumulativeStatistics {
+    /// Cumulative statistics created for this shard. When this shard reads statistics, it decrements `remaining messages`
```
SmaGMan, line 1283:
```suggestion
    /// Cumulative statistics created for this shard. When reader reads messages, it decrements `remaining messages`
```

collator/src/collator/types.rs
```diff
@@ -1279,6 +1280,9 @@ pub struct QueueStatisticsWithRemaning {
...
+    /// Another shard can decrements the statistics remaining only by using `update_processed_to_by_partitions`
```
SmaGMan, line 1284:
```suggestion
    /// Another shard stats can be decremented only by calling `update_processed_to_by_partitions`
```

collator/src/collator/messages_reader/mod.rs
drmick: `cumulative_stats_just_loaded` removed for enriched data because cumulative statistics is already has reduced `remaining stats`

collator/src/collator/do_collate/prepare.rs
```diff
@@ -148,6 +150,7 @@ impl Phase<PrepareState> {
                 reader_state: self.extra.reader_state,
                 anchors_cache: self.extra.anchors_cache,
                 is_first_block_after_prev_master: self.state.is_first_block_after_prev_master,
+                part_stat_ranges: self.state.part_stat_ranges.clone(),
```
SmaGMan, line 153: Why not to calc `part_stat_ranges` right here and not to pass it thru `ActualState.part_stat_ranges`?

drmick: This func isn't `async`

collator/src/collator/tests/collator_tests.rs
```diff
@@ -437,7 +443,7 @@ fn test_get_anchors_processing_info() {
     // will get anchors processing info from prev shard state
     let anchors_proc_info_opt = CollatorStdImpl::get_anchors_processing_info(
         &shard_id,
-        &mc_data,
+        &mc_data_stuff,
```
SmaGMan, line 446: `mc_data` was updated before but here you pass the old value

collator/src/collator/tests/collator_tests.rs
```diff
@@ -370,13 +370,19 @@ fn test_get_anchors_processing_info() {
         top_processed_to_anchor: 0,
         ref_mc_state_handle: tracker.insert(0),
         shards_processed_to_by_partitions: Default::default(),
+        prev_mc_block_id: None,
+    };
+
+    let mc_data_stuff = McDataStuff {
```
SmaGMan, line 376: Seems like you do not need McDataStuff in this test at all

collator/src/collator/mod.rs
```diff
@@ -986,7 +995,7 @@ impl CollatorStdImpl {
 
     fn get_anchors_processing_info(
         shard_id: &ShardIdent,
-        mc_data: &McData,
+        mc_data: &Arc<McData>,
```
SmaGMan, line 998: Better to return `mc_data: &McData,`

drmick:
> Might not be a good idea to store both the current and previous `McData` at the same time. This can prevent GC from clearing old state and it also holds references to some MC state subtrees longer than needed.
> 
> And the code/diff is now quite polluted with the usage of `mc_data_stuff.current` 🌚

We can store required prev mc data fields inside McData in special field `Option<McDataPrev>`
@SmaGMan

drmick:
> Might not be a good idea to store both the current and previous `McData` at the same time. This can prevent GC from clearing old state and it also holds references to some MC state subtrees longer than needed.
> 
> And the code/diff is now quite polluted with the usage of `mc_data_stuff.current` 🌚

If we remove unnecessary fields of previous mc data then we can store required prev mc data fields inside McData in special field `Option<McDataPrev>` and filling it only when we need it
@SmaGMan

Rexagon (approved):

PR #802, opened 2025-06-27
chore: fix clippy 1.88.0

0xdeafbeef:
Simple  `cargo clippy --all --all-targets --fix`
`match_on_vec_items` is now merged into `indexing_slicing`,  but it's to noisy


Rexagon:
`rustfmt` breaks on too long strings (>110 chars). Non-inlined format args were there to reduce the string length.

Please split large strings with ` \` to fix rustfmt issues

0xdeafbeef:
@Rexagon it was the the only one  line > 110 chars

Also there're bellion errors if you run cargo +nightly fmt [with this option](https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#error_on_line_overflow)

PR #836, opened 2025-07-21
refactor(storage): split block storage

0xdeafbeef:
BlockStorageConfig:

  archive_chunk_size removed because we don't split archives.
  split_block_tasks removed
   
  Not sure if we should set block chunk size in a rpc config, or simply set it as max size which can be sent via `tokio-codec`.

### COMPATIBILITY

Affected features:

Archive contents is byte by byte compatible with prev version, same with archive ids.

No migrations from previous version are provided, so this update should be done via resync.
Old rocks-db tables and migrations were removed to clean up the code.

Resync nodes one by one.

#### cpu
-18% irq cpu usage.
image
-44% io wait
image
-35% kernel space cpu usage
image
-15% user-space cpu usage
image

#### io

-80% io pressure, -65% io utilisation
image

-30% write traffic
image

-25% iops
image

-25% load avg
image

#### vm

-30% - 50% 
page faults

image

metrics comparison:
[metrics.html.zip](https://github.com/user-attachments/files/21353060/metrics.html.zip)

#### Manual Tests

pkill -9 of validator, was stable af. Re-syncs also worked flawlessly. 

## todo
- [] compare node metrics with 10m state + 10k transfers for 2 hours


0xdeafbeef:
10m state + 10k tps, new version time diff don't degrade
image
https://grafana.broxus.com/goto/h9Zmz9UNR?orgId=1 (2 nodes lost sync because of block size protection, i'll fix it in the next commit)

PR #840, opened 2025-07-24
perf(core): compute tail_len for gc beforahand while blocks are fresh

0xdeafbeef:
Before:
image
After:
image

Same with number of loads per second.
before:
image
after:
image


core/src/storage/block/tail_cache.rs
```diff
@@ -0,0 +1,176 @@
...
+
+            // If current_master > seqno and not finalized, it's too old
+            if let Some(current) = state.current_master {
+                if current.seqno > seqno {
+                    return None;
+                }
+            }
+        }
+
+        // Wait for finalization
```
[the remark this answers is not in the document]
0xdeafbeef: Nice catch! Fixed

drmick (approved):

core/src/storage/block/tail_cache.rs
```diff
@@ -0,0 +1,177 @@
...
+#[derive(Default)]
+struct InnerState {
+    current_master: Option<BlockId>,
+    current_max_tail_len: u32,
+    finalized: BTreeMap<u32, u32>, // seqno -> max_tail_len
+    waiters: HashMap<u32, Vec<oneshot::Sender<u32>>>, // seqno -> waiters
+}
+
+#[derive(Clone)]
+pub struct TailDiffCache {
```
SmaGMan, line 18: nit: better name is DiffTailCache

core/src/storage/block/tail_cache.rs
```diff
@@ -0,0 +1,177 @@
...
+                    }
+                }
+            }
+
+            // Start new round with master's tail_len as initial max
+            state.current_master = Some(block_id);
+            state.current_max_tail_len = tail_len;
+        } else {
+            // Update max if this shard has longer tail
+            state.current_max_tail_len = state.current_max_tail_len.max(tail_len);
```
SmaGMan, line 59: We should calculate max only among master and last shard block in each shard (currently we have only one) because only last shard block contains actual and relevant tail_len.

SmaGMan: Well. It is reasonable. But this way we can use this only for GC purposes.

core/src/storage/block/tail_cache.rs
```diff
@@ -0,0 +1,177 @@
...
+        let cache_clone = cache.clone();
+        let wait_task = tokio::spawn(async move { cache_clone.get(1).await });
+
+        // Shard blocks with different tail lengths
+        let shard1 = BlockStuff::new_with(ShardIdent::BASECHAIN, 100, |block| {
+            block.out_msg_queue_updates.tail_len = 20;
+        });
+        cache.update_block(&shard1);
+
+        let shard2 = BlockStuff::new_with(ShardIdent::new_full(0), 101, |block| {
```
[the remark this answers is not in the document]
0xdeafbeef: done

core/src/storage/block/tail_cache.rs
Rexagon: The current gc implementation is inaccurate towards the safer bound so we can live with it for now (mixing max diff of masterchain and other workchains can have an error of x32, and additional +32 might not be significant here)

PR #859, opened 2025-08-08
Reduce downloader's requests

k-morozov:

[None]

[None]

---

### COMPATIBILITY

[fully compatible]

[Not Required]

---

### PERFORMANCE IMPACT

[Expected impact]

Expect reducing amount of new requests when we have already downloading points.

#### Unit Tests

[Covered by:]

- linear_basic
- linear_already_downloading
- exponential_basic
- exponential_already_downloading

#### Network Tests

[No coverage]

#### Manual Tests

[Manual tests used]

---

**Notes/Additional Comments:**  

For resolving: https://github.com/broxus/tycho/issues/756

### DIFF
--- a/consensus/src/intercom/dependency/downloader.rs
+++ b/consensus/src/intercom/dependency/downloader.rs
@@ -43,29 +43,35 @@ struct DownloaderInner {
 }
 
 trait DownloadType: Send + 'static {
-    fn next_peers(attempt: u8, available_peers: usize, conf: &MempoolConfig) -> usize;
+    fn next_peers(
+        attempt: u8,
+        already_downloading: usize,
+        peer_count: &PeerCount,
+        conf: &MempoolConfig,
+    ) -> usize {
+        Self::max_downloads_per_attempt(attempt as usize, conf)
+            .min(peer_count.reliable_minority())
+            .saturating_sub(already_downloading)
+    }
+
+    fn max_downloads_per_attempt(attempt: usize, conf: &MempoolConfig) -> usize;
 }
 
 /// Exponential increase of peers to query
 struct ExponentialQuery;
 impl DownloadType for ExponentialQuery {
-    fn next_peers(attempt: u8, available_peers: usize, conf: &MempoolConfig) -> usize {
-        // result increases exponentially
+    fn max_downloads_per_attempt(attempt: usize, conf: &MempoolConfig) -> usize {
         let download_peers = conf.consensus.download_peers as usize;
-        download_peers
-            .saturating_mul(download_peers.saturating_pow(attempt as u32))
-            .min(available_peers)
+        download_peers.saturating_mul(download_peers.saturating_pow(attempt as u32))
     }
 }
 
 /// Linear increase of peers to query
 struct LinearQuery;
 impl DownloadType for LinearQuery {
-    fn next_peers(attempt: u8, available_peers: usize, conf: &MempoolConfig) -> usize {
+    fn max_downloads_per_attempt(attempt: usize, conf: &MempoolConfig) -> usize {
         let download_peers = conf.consensus.download_peers as usize;
-        download_peers
-            .saturating_add(download_peers.saturating_mul(attempt as usize))
-            .min(available_peers)
+        download_peers.saturating_add(download_peers.saturating_mul(attempt))
     }
 }
 
@@ -286,9 +292,15 @@ impl<T: DownloadType> DownloadTask<T> {
             .collect::<Vec<_>>();
         filtered.sort_unstable_by(|(_, ord_l), (_, ord_r)| ord_l.cmp(ord_r));
 
-        let count = T::next_peers(self.attempt, filtered.len(), self.ctx.conf());
+        let to_add = T::next_peers(
+            self.attempt,
+            self.downloading.len(),
+            &self.peer_count,
+            self.ctx.conf(),
+        )
+        .min(filtered.len());
 
-        for (peer_id, _) in &filtered[..count] {
+        for (peer_id, _) in &filtered[..to_add] {
             self.download_one(peer_id);
         }
 
@@ -439,6 +451,7 @@ impl<T: DownloadType> DownloadTask<T> {
         }
     }
 }
+
 impl DownloadCtx {
     fn meter_unreliable() {
         metrics::counter!("tycho_mempool_download_unreliable_responses").increment(1);
@@ -459,5 +472,81 @@ impl DownloadCtx {
 
         metrics::gauge!("tycho_mempool_download_depth_rounds")
             .set(self.download_max_depth(point_id.round));
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use anyhow::Result;
+
+    use super::*;
+    use crate::models::PeerCount;
+    use crate::test_utils::default_test_config;
+
+    fn test_impl<T: DownloadType>(
+        expected_to_add: &[usize],
+        already_downloading: usize,
+        peer_count: PeerCount,
+    ) -> Result<()> {
+        let mut attempt = 0;
+
+        let merged_conf = default_test_config();
+
+        for (step, _) in expected_to_add.iter().enumerate() {
+            let actual_peers =
+                T::next_peers(attempt, already_downloading, &peer_count, &merged_conf.conf);
+
+            anyhow::ensure!(
+                expected_to_add[step] == actual_peers,
+                "step {}, attempt {}, already_downloading {}, expected {}, got {}",
+                step,
+                attempt,
+                already_downloading,
+                expected_to_add[step],
+                actual_peers,
+            );
+
+            anyhow::ensure!(actual_peers + already_downloading <= peer_count.reliable_minority());
+
+            attempt = attempt.wrapping_add(1);
+        }
+
+        Ok(())
+    }
+
+    #[test]
+    fn linear_basic() -> Result<()> {
+        let expected_to_add = [2, 4, 6, 8, 10, 11, 11];
+        let already_downloading = 0;
+        let peer_count = PeerCount::try_from(30)?;
+
+        test_impl::<LinearQuery>(&expected_to_add, already_downloading, peer_count)
+    }
+
+    #[test]
+    fn linear_already_downloading() -> Result<()> {
+        let expected_to_add = [0, 0, 2, 4, 6, 7, 7, 7];
+        let already_downloading = 4;
+        let peer_count = PeerCount::try_from(30)?;
+
+        test_impl::<LinearQuery>(&expected_to_add, already_downloading, peer_count)
+    }
+
+    #[test]
+    fn exponential_basic() -> Result<()> {
+        let expected_to_add = [2, 4, 8, 11, 11];
+        let already_downloading = 0;
+        let peer_count = PeerCount::try_from(30)?;
+
+        test_impl::<ExponentialQuery>(&expected_to_add, already_downloading, peer_count)
+    }
+
+    #[test]
+    fn exponential_already_downloading() -> Result<()> {
+        let expected_to_add = [0, 0, 4, 7, 7];
+        let already_downloading = 4;
+        let peer_count = PeerCount::try_from(30)?;
+
+        test_impl::<ExponentialQuery>(&expected_to_add, already_downloading, peer_count)
     }
 }


Mododo (changes requested):

Mododo (approved):

consensus/src/intercom/dependency/downloader.rs
```diff
@@ -461,3 +490,78 @@ impl DownloadCtx {
...
+    fn linear_basic() -> Result<()> {
+        let expected_to_add = [2, 4, 6, 8, 10, 11, 11];
+        let already_downloading = 0;
+
+        test_impl::<LinearQuery>(&expected_to_add, already_downloading)
+    }
+
+    #[test]
+    fn linear_already_downloading() -> Result<()> {
+        let expected_to_add = [0, 0, 2, 4, 6, 8, 10, 11];
```
Mododo, line 546: let expected_to_add = [0, 0, 2, 4, 6, 7, 7, 7];

Mododo (changes requested):

consensus/src/intercom/dependency/downloader.rs
k-morozov: Updated

0xdeafbeef:
> > so subtracting them again is double accounting
> 
> @0xdeafbeef Not exactly. Already filtered and currently downloading are different sets. However, our goal is to reduce network usage and we use `next_peers` to calculate the amount of possible new requests.
> 
> For example: filtered peers: 10 items and curently downloading 10 as well. On current step we are going to process only 10 items. So wihtout subtracting we will add 10 new request. It's not that we want. In this case we don't want to make any reqeusts.

I don't think your example holds. 

Try this one:
 ```
 fn next_peers(
        attempt: u8,
        available_peers: usize,
        already_downloading: usize,
        peer_count: &PeerCount,
        conf: &MempoolConfig,
    ) -> usize {
        // result increases exponentially
        let download_peers = conf.consensus.download_peers as usize;
        download_peers
            .saturating_mul(download_peers.saturating_pow(attempt as u32))
            .min(available_peers)
            .min(peer_count.reliable_minority())
            .saturating_sub(already_downloading)
    }
    ```
    
    attempt 2, available_peers 5, already_downloading 5.
 

` .saturating_mul(download_peers.saturating_pow(attempt as u32)) ` is 8
 The function calculates `min(8, 5) - 5`  =  0 (i assume reliable_minority is bigger then 5, if lower than it will anyway give 0) . We start no new downloads, even though we should have started 3 to reach our target of 8.
 
 So we have 5 free peers and restrict ourselves from downloading instead of increasing number of parallel queries. Maybe I've incorrectly got  intention of this

k-morozov:
> `.saturating_mul(download_peers.saturating_pow(attempt as u32))` is 8 The function calculates `min(8, 5) - 5` = 0 (i assume reliable_minority is bigger then 5, if lower than it will anyway give 0) . We start no new downloads, even though we should have started 3 to reach our target of 8.
> 
> So we have 5 free peers and restrict ourselves from downloading instead of increasing number of parallel queries. Maybe I've incorrectly got intention of this

In this case we can not dowload greater than `reliable_minority` - 5. So if we have already dowloading 5 the `next_peers` must return 0. If `reliable_minority` would for example 10 the `next_peers` must return 3. It works well.

> We start no new downloads, even though we should have started 3 to reach our target of 8
Seems that this thought should be discusses in the [issue](https://github.com/broxus/tycho/issues/756). Right now was decided to limit by `reliable_minority`. I assume the limit makes seanse and I know no reason don't  use it.

0xdeafbeef:
> > `.saturating_mul(download_peers.saturating_pow(attempt as u32))` is 8 The function calculates `min(8, 5) - 5` = 0 (i assume reliable_minority is bigger then 5, if lower than it will anyway give 0) . We start no new downloads, even though we should have started 3 to reach our target of 8.
> > So we have 5 free peers and restrict ourselves from downloading instead of increasing number of parallel queries. Maybe I've incorrectly got intention of this
> 
> In this case we can not dowload greater than `reliable_minority` - 5. So if we have already dowloading 5 the `next_peers` must return 0. If `reliable_minority` would for example 10 the `next_peers` must return 3. It works well.
> 
> > We start no new downloads, even though we should have started 3 to reach our target of 8
> 
> Seems that this thought should be discusses in the [issue](https://github.com/broxus/tycho/issues/756). Right now was decided to limit by `reliable_minority`. I assume the limit makes seanse and I know no reason don't use it.

You can use `filtered.len() + self.downloading.len()` as `available_peers` 

It will give you 0 peers in your example -  `min(8, 20) - 10 == 0`

and in my case
`attempt 2, available_peers becomes 10 (5 idle + 5 busy), already_downloading 5.`
 ```
 .saturating_mul(download_peers.saturating_pow(attempt as u32)) = 8 as before
 min(8, 10) - 5 = 3.
```

**Comment by 0xdeafbeef** [comment]
> > You can use `filtered.len() + self.downloading.len()` as `available_peers`
> 
> is it make seance?
> 
> Show me the tests pls that could explain you idea and show the different from original code and my solution. I found it more usefull for our discussion.


### The Universe of Peers: `self.undone_peers`

First, let's establish the "universe" we're working with.

*   `self.undone_peers`: This is a `FastHashMap<PeerId, PeerStatus>`. It contains every peer that is still a candidate for being queried for the point. A peer is removed from this map only when it successfully provides the point or when it provides a `DefinedNone` response (and is considered reliable).

At the moment `download_random` is called, this universe of `undone_peers` can be partitioned based on the `PeerStatus` of each peer. The two most important fields in `PeerStatus` for this function are:
*   `state: PeerState`: Is the peer considered reachable (`Resolved`) or not?
*   `is_in_flight: bool`: Do we have an active, outstanding query to this peer right now?

### The Sets: `filtered` and `self.downloading`

The code partitions the `Resolved` peers from the universe into two distinct groups.

#### 1. `filtered` (The Idle Peers)
This is a temporary `Vec` created inside `download_random`.

*   **What it is:** It is the set of all peers that are **both reachable AND currently idle**.
*   **How it's created:**
    ```rust
    let mut filtered = self
        .undone_peers
        .iter()
        // The two conditions that define this set:
        .filter(|(_, p)| p.state == PeerState::Resolved && !p.is_in_flight)
        .map(|(peer_id, _)| *peer_id) // simplified for explanation
        .collect::<Vec<_>>();
    ```
*   **Its size `filtered.len()`:** This number represents the **count of available peers we can immediately send a new request to.**

#### 2. `self.downloading` (The Busy Peers)
This is a `FuturesUnordered` collection that is a member of the `DownloadTask` struct.

*   **What it is:** It holds the active `Future` for each peer that has an **in-flight network request**. When a request completes, its `Future` is resolved and removed from this set.
*   **How it's populated:** A new future is added via `self.downloading.push(...)` every time we start a new query in `download_one`.
*   **Its size `self.downloading.len()`:** This number represents the **count of peers we are currently waiting on a response from.**

### Relationship: Which is a subset of which?

**`filtered` and `self.downloading` are disjoint sets.**

A peer cannot be in both sets at the same time.
*   If a peer is in `filtered`, its `is_in_flight` status must be `false`.
*   If a peer is represented in `self.downloading`, its `is_in_flight` status is `true`.

Their union represents the total pool of resolved (reachable) peers.

**`total_resolved_peers = filtered.len() + self.downloading.len()`**

---

### Tracing the Numbers: A Concrete Example

Let's trace where each argument to `T::next_peers` comes from, using a hypothetical scenario.

**Scenario:**
*   Total validators for the round: **30** (`peer_count` is based on this)
*   Current attempt number: **2** (`self.attempt` = 2)
*   Peers with active, in-flight requests: **5**
*   Peers that are reachable but idle: **12**
*   Peers that are unreachable (e.g., `PeerState::Unresolved`): 13

**Argument Trace for `T::next_peers(...)`**

| Argument | Value | Origin and Meaning |
| :--- | :--- | :--- |
| `self.attempt` | `2` | A simple counter in `DownloadTask`. It's incremented after each `tick`, driving the linear/exponential increase in the number of peers to query. |
| `total_resolved_peers` **(The Fix)** | `17` | This is **`filtered.len() + self.downloading.len()`** (`12 + 5`). It represents the **total pool of currently reachable peers**. This is the correct number to use as the upper bound on how many peers we *could* be talking to. |
| `self.downloading.len()` | `5` | The number of futures in the `self.downloading` set. It represents the **number of peers already busy** with a query. The `next_peers` logic will subtract this from its target. |
| `&self.peer_count` | `PeerCount(30)` | Initialized once when the `DownloadTask` starts. It represents the **total size of the validator set**. It's used to calculate security thresholds like `reliable_minority()`, ensuring we don't query more peers than necessary for BFT safety. |
| `self.ctx.conf()` | `MempoolConfig` | The static configuration passed into the task. It contains values like `download_peers` (e.g., `2`), which is the base for the linear/exponential calculation. |

#### How the Calculation Works (with `ExponentialQuery`)

1.  **Calculate the raw target:**
    `target = download_peers * download_peers.pow(attempt)`
    `target = 2 * 2.pow(2) = 2 * 4 = 8`

2.  **Clamp the target:** The target is clamped by the available resources and security limits.
    `clamped_target = target.min(total_resolved_peers).min(peer_count.reliable_minority())`
    Let's say `reliable_minority()` is 11.
    `clamped_target = 8.min(17).min(11) = 8`

3.  **Calculate new peers to add:** Subtract the peers that are already busy.
    `to_add = clamped_target.saturating_sub(already_downloading)`
    `to_add = 8 - 5 = 3`

**Result:** The function returns `3`. The system will now start 3 new downloads from the `filtered` (idle) peer list, bringing the total number of in-flight requests to `5 + 3 = 8`, perfectly matching the target.

#### Why the Original Code Was Wrong

The original code would have passed `filtered.len()` (which is `12`) as `available_peers`.
`clamped_target = 8.min(12).min(11) = 8`
`to_add = 8 - 5 = 3`

In this *specific* example, the result is the same. But let's change one number: assume we want to query **15** peers.

*   **Correct Logic:** `clamped_target = 15.min(17).min(11) = 11`. `to_add = 11 - 5 = 6`.
*   **Buggy Logic:** `clamped_target = 15.min(12).min(11) = 11`. `to_add = 11 - 5 = 6`.

Still works. Let's try another. Target **15**, `filtered.len()` is **10**, `downloading.len()` is **4**. Total resolved is **14**.

*   **Correct Logic:** `clamped_target = 15.min(14).min(11) = 11`. `to_add = 11 - 4 = 7`.
*   **Buggy Logic:** `clamped_target = 15.min(10).min(11) = 10`. `to_add = 10 - 4 = 6`.

Here we see the bug: the old logic would only add 6 peers because it was incorrectly capped by the number of *idle* peers (`10`), failing to add the 7th peer even though a resource was available. The fix allows it to correctly calculate and add 7 peers.

**Comment by Mododo** [comment]
downloading2



given
1) `Available = 3F+1 - Current_Downloads - Unresolved`
2) `Current_Downloads + To_start <=  MAX_Downloads <= 1F+1`

current formula is
`To_start = MIN (Available, MAX_Downloads) - Current_Downloads`

while it should really be

`To_start = MIN (Available, MAX_Downloads - Current_Downloads)`

-------------

Please keep the discussion more concise and close to the matter. It could have been a short bug fix instead of the long bug report.

File: consensus/src/intercom/dependency/downloader.rs [hunk 2]
```diff
@@ -286,9 +294,15 @@ impl<T: DownloadType> DownloadTask<T> {
...
         filtered.sort_unstable_by(|(_, ord_l), (_, ord_r)| ord_l.cmp(ord_r));
 
-        let count = T::next_peers(self.attempt, filtered.len(), self.ctx.conf());
+        let to_add = T::next_peers(
+            self.attempt,
+            filtered.len(),
+            self.downloading.len(),
+            &self.peer_count,
+            self.ctx.conf(),
+        );
```
Mododo, line 303: just remove `available_peers` arg from `next_peers()` and use
`let to_add = T::next_peers(...).min(filtered.len());`
you'll notice how the code becomes more simple, and also all tests pass

Mododo (changes requested):

consensus/src/intercom/dependency/downloader.rs
Mododo: also the second and the last test cases become equal, so just remove one of them

k-morozov: I suggest leaving `available_peers` within `next_peers`. Otherwise evaluation's logic will be in different places instead have only one function for calculation limit. In addition, the function will not be self-sufficient because you must always remember that you need to add a filter.

k-morozov: Updated

PR #865, opened 2025-08-12
feature(collator): add support `mc_block_max_interval_ms`

drmick:


collator/src/manager/mod.rs
```diff
@@ -2682,151 +2692,169 @@ where
...
-                );
-                false
+            fact.has_any_sample = true;
+
+            for s in &state.last_imported_samples {
+                // capture first imported anchor strictly after masterchain ct
+                if fact.after_mc_ct.is_none() && s.ct > mc_ct {
+                    fact.after_mc_ct = Some(s.ct);
+                }
+                if fact.forced_ct.is_none() && s.forced {
```
SmaGMan, line 2750: Seems like we should demand every event ct to be > last mc block ct. Maybe it will be clearer to use a filter on the iterator over `state.last_imported_samples`.

collator/src/manager/types.rs
```diff
@@ -430,3 +439,47 @@ pub struct HandledBlockFromBcCtx {
...
+    pub chain_time: u64,
+    pub force_mc_block: ForceMasterCollation,
+    pub trigger_shard_block_id_opt: Option<BlockId>,
+    pub has_collated_block_after_prev_master: bool,
+    pub mc_block_min_interval_ms: u64,
+    pub mc_block_max_interval_ms: u64,
+}
+
+#[derive(Debug)]
+pub struct DetectCollationCtx {
```
SmaGMan, line 455: Is used only inside `manager/mod.rs`, so better to move there.

collator/src/manager/mod.rs
```diff
@@ -683,68 +687,65 @@ where
...
-
         tracing::debug!(target: tracing_targets::COLLATION_MANAGER,
             next_step = ?next_step,
             "detected next collation step",
         );
 
         match next_step {
             NextCollationStep::CollateMaster(next_mc_block_chain_time) => {
-                if !shard_id.is_masterchain() {
+                if !ctx.shard_id.is_masterchain() {
```
SmaGMan, line 730: Better to destruct context object at the beginning to prevent too many code line changes.
```
let CollationStepContext { shard_id, ... } = ctx;
```

collator/src/manager/mod.rs
```diff
@@ -2693,120 +2682,124 @@ where
...
-            for s in &state.last_imported_samples {
-                // capture first imported anchor strictly after masterchain ct
-                if fact.after_mc_ct.is_none() && s.ct > mc_ct {
-                    fact.after_mc_ct = Some(s.ct);
-                }
-                if fact.forced_ct.is_none() && s.forced {
+            for s in &state.last_imported_anchor_events {
+                fact.has_any_anchor_event = true;
+
+                if s.mc_forced {
```
SmaGMan, line 2732: Why do you not check that `fact.forced_ct.is_none()`. Without it will contain last, not first, as intended.

collator/src/manager/mod.rs
```diff
@@ -683,68 +687,65 @@ where
...
-            self.active_collation_sessions
-                .read()
-                .keys()
-                .cloned()
-                .collect(),
-            shard_id,
-            chain_time,
-            force_mc_block,
-            mc_block_min_interval_ms,
+            DetectCollationCtx::new(
```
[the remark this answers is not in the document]
drmick: Less code and, according to clippy, not so many fields to avoid functions

collator/src/manager/mod.rs
drmick: But desеcructor add more lines than using changes with `ctx.`

```
       let CollationStepContext {
            shard_id,
            chain_time,
            force_mc_block,
            has_collated_block_after_prev_master,
            mc_block_min_interval_ms,
            mc_block_max_interval_ms,
            prev_mc_block_id,
            trigger_shard_block_id_opt,
        } = ctx;
```

SmaGMan (approved):

PR #870, opened 2025-08-19
ci: add doc test

drmick:


util/src/config.rs
```diff
@@ -6,6 +6,7 @@ pub use tycho_util_proc::PartialConfig;
 ///
 /// By default all fields are "non-important", e.g.:
 /// ```
+/// use tycho_util_proc::PartialConfig;
```
Rexagon, line 9: You can use
```rust
# use tycho_util_proc::PartialConfig;
#[derive(Default, PartialConfig)]
struct EmptyConfig {
    ...
}
```

To hide the import while preserving the clean output of rendered docs

Rexagon (approved):

PR #900, opened 2025-09-12
Refactor/consensus/download types

Mododo:
usage of updated consensus config https://github.com/broxus/tycho-types/pull/55

Yes

Consensus config param `download_tasks: u16` moved to mempool node config param `max_download_tasks: u16`. New default value 250 replaces previous 260.

Consensus config (param 29) field `download_tasks: u16` was replaced with `min_sign_attempts: u8` and `download_peer_queries: u8` . Compared to v0.2.12, config value stays of the same size, as `broadcast_retry_attempts` was removed from the end of the struct (was added in https://github.com/broxus/tycho/pull/867 with new tag which is also removed)

---

### COMPATIBILITY

Affected features:

- State. Blockchain Config
- Mempool. Consensus Config

New genesis round and time __must__ be applied to global config, because `download_peer_queries` affects overlay id, and new overlay id does not support history from the old genesis.

At least a couple of minutes before network upgrade, consensus config param `download_tasks` should be set to 778, so that it will be parsed as `min_sign_attempts` = 3 and `download_peer_queries` = 10 by new version. Outdated nodes should not experience high load before upgrade, so that amount of simultaneous download tasks stays low and does not hit 778.  
Also config may be changed after network upgrade: until then, `download_tasks` = 260 will be parsed into  `min_sign_attempts` = 1 and `download_peer_queries` = 4. It is a valid and operable config, but a rounds rate will be a bit higher than desirable, and high load should be avoided too before config change.

### PERFORMANCE IMPACT

No impact expected

---

#### Unit Tests

Covered by config serde in `tycho-types`

#### Network Tests

No special coverage

#### Manual Tests

transfers-30k
network upgrade

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4206,7 +4206,7 @@ dependencies = [
 [[package]]
 name = "tycho-types"
 version = "0.2.1"
-source = "git+https://github.com/broxus/tycho-types.git?rev=715769a16fc4497a60865948eb15be928f974446#715769a16fc4497a60865948eb15be928f974446"
+source = "git+https://github.com/broxus/tycho-types.git?rev=35fea56d604d00cc0f4b4e967d6326ff0793ef26#35fea56d604d00cc0f4b4e967d6326ff0793ef26"
 dependencies = [
  "ahash",
  "anyhow",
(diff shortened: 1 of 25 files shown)


SmaGMan:
Was it tested in devnet: the config change after network upgrade? Looks like it should be the main upgrade scenario.

PR #904, opened 2025-09-17
fast sync diffs applied upto

SmaGMan:
It may be a situation when the message queue was committed after block validation, but the block was not applied before node restarts. In this case, we may get a panic after a restart during queue restore because we first take the last collated/synced_to mc block as the "queue applied to" border. However, we should take the last committed instead if it is above.

---

[None]

[None]

---

### COMPATIBILITY

[fully compatible]

[Not Required]

---

### PERFORMANCE IMPACT

[No impact expected]

---

#### Unit Tests

Covered by:
- test_queue_restore_on_sync

#### Network Tests

[No coverage]

#### Manual Tests

Manual tests used
- transfers 30k with many network start/stop when messages queues are not empty

### DIFF
--- a/collator/src/manager/mod.rs
+++ b/collator/src/manager/mod.rs
@@ -1764,13 +1764,25 @@ where
         &self,
         last_collated_mc_block_id: Option<BlockId>,
     ) -> Result<Option<BlockId>> {
-        match self.get_top_mc_block_id_for_next_collation(last_collated_mc_block_id) {
-            // return last collated if it exists or last "synced to"
-            Some(block_id) => Ok(Some(block_id)),
-            // otherwise we just started and do not have stored last collated or "synced to"
-            // in this case we read last master block id comitted into queue
-            None => self.mq_adapter.get_last_commited_mc_block_id(),
-        }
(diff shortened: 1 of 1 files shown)


collator/src/storage/transaction.rs
```diff
@@ -66,39 +65,21 @@ impl InternalQueueTransaction {
...
-        commit_pointers: &FastHashMap<ShardIdent, (QueueKey, u32)>,
+        commit_pointers: FastHashMap<ShardIdent, (QueueKey, u32)>,
     ) -> Result<()> {
         let commit_pointers_cf = self.db.internal_message_commit_pointer.cf();
 
-        for (&shard_ident, (queue_key, seqno)) in commit_pointers.iter() {
+        for (shard_ident, (queue_key, seqno)) in commit_pointers {
             let key = CommitPointerKey { shard_ident }.to_vec();
 
-            // Get the old value if it exists
```
drmick, line 78: Why do you removed this logic but didn't add to caller in same commit? Won't this change the logic of the node's work and cause errors?

PR #912, opened 2025-09-21
fix(core): init allowed workchains if boot is not cold

Mododo:

None

None

---

### COMPATIBILITY

Full 

Not Required

---

### PERFORMANCE IMPACT

No impact expected

---

#### Unit Tests

No coverage

#### Network Tests

No coverage

#### Manual Tests

`transfers-30k.sh --continue`


Rexagon (approved):

Rexagon:
Btw - better move this allowed workchains management logic somewhere into the strider since in theory new workchains can be added/removed dynamically from the config. And receiving messages can also be dynamically toggled for them.

Since this is just an "early-discard" optimization, there might not even be a need for this feature inside deserialization. In any case, in the future there will be another layer of full-fledged validation with execution where it will be properly discarded.

PR #931, opened 2025-10-08
feat(cli): add check cells db tool

serejkaaa512:

[None] 

[None] 

---

### COMPATIBILITY

[fully compatible] 

[Not Required] 

### PERFORMANCE IMPACT

[No impact expected] 

---

#### Unit Tests

[No coverage]

#### Network Tests

[No coverage]

#### Manual Tests

Run local node, run tool to check consistency

### DIFF
--- a/cli/src/cmd/tools/check_cells_db.rs
+++ b/cli/src/cmd/tools/check_cells_db.rs
@@ -1,0 +1,195 @@
+use std::path::{Path, PathBuf};
+
+use anyhow::{Context, Result};
+use bumpalo_herd::Herd;
+use bytesize::ByteSize;
+use clap::Parser;
+use tempfile::TempDir;
+use tycho_block_util::dict::split_aug_dict_raw;
+use tycho_core::storage::{CoreStorage, CoreStorageConfig};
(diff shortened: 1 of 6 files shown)


cli/src/cmd/tools/check_cells_db.rs
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    println!(
+        "Copying cells database from {} to {}...",
+        cells_src.display(),
+        cells_dst.display()
+    );
+    copy_dir_recursive(&cells_src, &cells_dst)?;
+
+    // 2. Open the copied database and remove all states
```
SmaGMan: We should use existing functions to remove states. We have `ShardStateStorage.remove_outdated_states` - it finds `TopBlocks` then perform deletion. We can split `remove_outdated_states` on 2 functions: 1st finds `TopBlocks`, 2nd - removes state upto `TopBlocks`. Then we can use the 2nd func to remove all state in current tool. We just need to pass the highest top blocks to it.

SmaGMan (approved):

PR #937, opened 2025-10-14
Reduce states writes

pashinov:

Yes

Added .core_storage.store_shard_state_step: u32 with default value 2.

None

---

### COMPATIBILITY

Fully compatible

Not Required

---

### PERFORMANCE IMPACT

States GC stopped lagging -> cells cache stopped growing -> network degradation fixed. TPS increased by 10-15%. Storage timings a little decreased.

---

### TESTS

Yes

#### Manual Tests

Deploy 10M accounts then run 20k transfers

image
image

### DIFF
--- a/core/src/storage/shard_state/mod.rs
+++ b/core/src/storage/shard_state/mod.rs
@@ -1,4 +1,3 @@
-use std::collections::BTreeMap;
 use std::fs::File;
 use std::io::Cursor;
 use std::num::NonZeroU8;
@@ -46,11 +45,12 @@ pub struct ShardStateStorage {
 
     accumulated_per_shard: parking_lot::Mutex<FastHashMap<ShardIdent, ShardAccumulator>>,
 
-    shard_states_cache: FastDashMap<ShardIdent, BTreeMap<u32, ShardStateStuff>>,
+    shard_states_cache: FastDashMap<ShardIdent, FastHashMap<u32, ShardStateStuff>>,
 
     shard_split_depth: u8,
     new_cells_threshold: usize,
     store_shard_state_step: NonZeroU8,
+    states_cache_force_reload_interval: NonZeroU8,
 }
 
 impl ShardStateStorage {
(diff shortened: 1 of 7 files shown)


collator/src/collator/mod.rs
```diff
@@ -666,8 +667,12 @@ impl CollatorStdImpl {
                     // get last store task
                     let last_task = self.store_new_state_tasks.pop().expect("shouldn't happen");
 
+                    let store_state_step = self.state_node_adapter.store_shard_state_step();
+
                     // if it is finished, then we can just reload prev state
-                    if last_task.store_new_state_task.is_finished() {
+                    if last_task.store_new_state_task.is_finished()
+                        && last_task.block_id.seqno.is_multiple_of(store_state_step)
```
SmaGMan: Maybe it will be better to use more clear helper, e.g. `self.state_node_adapter.check_should_store_state(..block_id.seqno)`?

core/src/storage/shard_state/mod.rs
```diff
@@ -130,8 +144,48 @@ impl ShardStateStorage {
...
+        // TODO: Move node_state to ShardStateStorage and check handle.ref_by_mc_seqno() == node_state.load_zerostate_id().seqno ???
+        let mut should_store = handle.is_masterchain()
+            || hint.is_top_block == Some(true)
+            || handle
+                .id()
+                .seqno
+                .is_multiple_of(self.store_shard_state_step.get() as u32);
+
+        if !should_store {
+            let mut guard = self.accumulated_per_shard.lock().await;
```
Rexagon: Use sync mutex here

core/src/storage/shard_state/mod.rs
```diff
@@ -130,8 +144,48 @@ impl ShardStateStorage {
...
+            let acc = guard.entry(handle.id().shard).or_default();
+
+            // prevents double counting
+            if acc.blocks.insert(handle.id().seqno) {
```
Rexagon: We should ensure that states are stored sequentially (at the collator level at least, because strider is always sequential).

core/src/storage/shard_state/mod.rs
```diff
@@ -246,7 +304,7 @@ impl ShardStateStorage {
     // knowing its `min_ref_mc_seqno` which can only be found out by
     // parsing the state. Creating a "Brief State" struct won't work either
     // because due to model complexity it is going to be error-prone.
-    pub async fn load_state(
+    pub async fn direct_load_state(
```
Rexagon: nit: load_state_direct

PR #947, opened 2025-11-04
fix(ci): check built metrics chart

Mododo:
Build dashboard to find metrics which names in dashboard are defined as concatenation.

Previously `check-metrics.py` was able to generate false positives in case the name is defined as concatenation in dashboard thus not found during check.

This removes the ability to name metric somewhere in dashboard comments to suppress warning.

None

None

---

### COMPATIBILITY

Full

Not Required

### PERFORMANCE IMPACT

No impact expected

---

#### Unit Tests

No coverage

#### Network Tests

No coverage

#### Manual Tests

run script


0xdeafbeef (approved):
lgtm, squash my changes if you are ok with it

Issue #932, opened 2025-10-08
Add option to `gen-account` to specify balance in extra currencies


Rexagon:
We should be able to generate accounts with some additional extra currencies to easily test some things without changing the config.

PR #961, opened 2025-11-20
feat(cli): add `--extra-balance` flag to specify a full CC

Rexagon:
Example:
```
tycho tool gen-account wallet \
  --pubkey 28982b175eba4a48c661f9d44e9a44e1a427b74f464673574f109127e8c3d3c6 \
  --balance 1000.100 \
  --extra-balance 1:1000,2:100500 \
  --extra-balance 3:123000000000000000000000000000000000000000000000000
```

Fixes https://github.com/broxus/tycho/issues/932

### DIFF
--- a/cli/src/cmd/tools/gen_account.rs
+++ b/cli/src/cmd/tools/gen_account.rs
@@ -1,7 +1,13 @@
+use std::collections::BTreeMap;
+use std::str::FromStr;
+
 use anyhow::{Context, Result};
 use tycho_crypto::ed25519;
-use tycho_types::models::{Account, AccountState, OptionalAccount, StateInit, StdAddr};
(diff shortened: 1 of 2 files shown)


cli/src/cmd/tools/gen_account.rs
```diff
@@ -379,9 +382,62 @@ impl GiverBuilder {
...
+        account.balance = self.balance;
         account.storage_stat.used = compute_storage_used(&account)?;
 
         Ok((address, account))
     }
 }
+
+#[derive(clap::Args)]
+struct BalanceArgs {
+    /// Initial balance of the wallet.
```
0xdeafbeef: Just `/// Initial balance` cause it's used with giver, not only wallets

PR #969, opened 2025-12-03
feat(collator): pass prev mc block id to executor

Rexagon:
An updated VM version has new opcodes and also a logic for building a tuple with prev blocks. For slashing support we only need the single entry of a previous masterchain block id.

It also brings support for a new `SignatureDomain` capability.


0xdeafbeef (approved):
lgtm

MrWad3r (approved):

PR #975, opened 2025-12-12
Ability to create generic forked state and run network using prepared state

MrWad3r:

[None]

[None]

---

### COMPATIBILITY

Fully compatible

Affected features:

- [State]
- [Persistent State]
- [Queue. Diff]
- [Collator]

Removed hardcoded checks for seqno == 0 in the collator and message queue logic. These checks were replaced with comparisons against zerostate_id.seqno, ensuring correct behavior when starting execution from a non-zero zerostate sequence number.

[Not Required] 

---

### PERFORMANCE IMPACT

[No impact expected]

---

#### Unit Tests

[Covered by:] 

Previous existing collator and queue tests

#### Network Tests
[No coverage]

#### Manual Tests

1. Start a local network consisting of three validator nodes. Allow the network to run until multiple masterchain and shard blocks are produced.
2. Stop the network. Using the new CLI command, generate zerostate files corresponding to a specified masterchain block.
3. Decompress the generated zerostate files using zstd.
4. Restart the network using the decompressed zerostate files as the initial state. The network must resume execution starting from the specified masterchain zerostate block.

Additionally, a Python virtual environment (venv) was used to set up tooling for token transfers from the giver account to a generic account. All transfer operations completed successfully, confirming correct network behavior.

### DIFF
--- a/collator/src/collator/do_collate/finalize.rs
+++ b/collator/src/collator/do_collate/finalize.rs
@@ -62,6 +62,7 @@ pub struct FinalizeBlockContext {
     pub processed_upto: ProcessedUptoInfoStuff,
     pub diff_tail_len: u32,
     pub block_serializer_cache: BlockSerializerCache,
+    pub zerostate_id: ZerostateId,
 }
 
 impl Phase<FinalizeState> {
@@ -102,11 +103,7 @@ impl Phase<FinalizeState> {
         let mut other_updated_top_shard_diffs_info = FastHashMap::default();
 
         for top_block_id in top_other_updated_shard_blocks_ids.iter() {
-            // FIXME: Use `zerostate_id.seqno` here, but we need to check
-            // against `ref_by_mc_seqno`.
-            if top_block_id.seqno == 0
-                || top_block_id.is_masterchain() && top_block_id.seqno <= zerostate_id.seqno
-            {
+            if top_block_id.seqno == 0 || self.state.mc_data.block_id.seqno <= zerostate_id.seqno {
                 continue;
             }
 
@@ -178,6 +175,7 @@ impl Phase<FinalizeState> {
             processed_upto,
             diff_tail_len,
             block_serializer_cache,
+            zerostate_id,
         } = ctx;
 
         let accounts_split_depth = collator_config.accounts_split_depth;
@@ -317,6 +315,7 @@ impl Phase<FinalizeState> {
                 prev_state,
                 prev_processed_to_anchor,
                 collator_config,
+                zerostate_id,
             )?;
             self.state
                 .collation_data
@@ -795,6 +794,7 @@ impl Phase<FinalizeState> {
         prev_state: &ShardStateStuff,
         prev_processed_to_anchor: u32,
         collator_config: Arc<CollatorConfig>,
+        zerostate_id: ZerostateId,
     ) -> Result<(McStateExtra, u32)> {
         // 1. update config params and detect key block
         let prev_state_extra = prev_state.state_extra()?;
@@ -962,7 +962,7 @@ impl Phase<FinalizeState> {
         });
 
         // 6. update prev_blocks (add prev block's id to the dictionary)
-        let prev_is_key_block = collation_data.block_id_short.seqno == 1 // prev block is a keyblock if it is a zerostate
+        let prev_is_key_block = collation_data.block_id_short.seqno == zerostate_id.seqno + 1 // prev block is a keyblock if it is a zerostate
             || prev_state_extra.after_key_block;
         let mut prev_blocks = prev_state_extra.prev_blocks.clone();
         let prev_blk_ref = BlockRef {
--- a/collator/src/internal_queue/queue.rs
+++ b/collator/src/internal_queue/queue.rs
@@ -24,7 +24,7 @@ use crate::internal_queue::types::stats::{
     AccountStatistics, DiffStatistics, SeparatedStatisticsByPartitions,
 };
 use crate::storage::models::DiffInfo;
-use crate::types::TopBlockId;
+use crate::types::TopBlockIdUpdated;
 use crate::{internal_queue, tracing_targets};
 
 #[derive(Debug, Serialize, Deserialize)]
@@ -93,7 +93,7 @@ where
     /// Commit diffs to the state and update GC
     fn commit_diff(
         &self,
-        mc_top_blocks: &[TopBlockId],
+        mc_top_blocks: &[TopBlockIdUpdated],
         partitions: &FastHashSet<QueuePartitionIdx>,
     ) -> Result<()>;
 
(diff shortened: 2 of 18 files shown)


cli/src/cmd/tools/prepare_hardfork.rs
```diff
@@ -0,0 +1,313 @@
...
+        .await?;
+
+        let storage = CoreStorage::open(ctx, CoreStorageConfig::default()).await?;
+
+        let Some(block_id) = storage
+            .shard_state_storage()
+            .load_mc_block_id(self.mc_seqno)?
+        else {
+            anyhow::bail!("mc block not found");
+        };
```
0xdeafbeef: nit: .with_context(||"format!("mc block not found. Seqno: {} ...

collator/src/manager/blocks_cache.rs
```diff
@@ -1113,14 +1123,14 @@ impl ReceivedBlockContext {
...
                 processed_upto: Default::default(),
             });
         }
 
-        let Some(ref_by_mc_seqno) = state_node_adapter.get_ref_by_mc_seqno(block_id).await? else {
-            bail!("block not found: {block_id}");
-        };
+        // let Some(ref_by_mc_seqno) = state_node_adapter.get_ref_by_mc_seqno(block_id).await? else {
+        //     bail!("block not found: {block_id}");
+        // };
```
0xdeafbeef: Why is it commented out?

core/src/storage/persistent_state/shard_state/writer.rs
```diff
@@ -29,6 +30,14 @@ impl<'a> ShardStateWriter<'a> {
...
+        PathBuf::from(name).with_extension(Self::FILE_EXTENSION)
+    }
+
+    pub fn temp_file_name_str(name: &str) -> PathBuf {
+        PathBuf::from(name).with_extension(Self::FILE_EXTENSION_TEMP)
+    }
+
     pub fn file_name(block_id: &BlockId) -> PathBuf {
         PathBuf::from(block_id.to_string()).with_extension(Self::FILE_EXTENSION)
     }
```
0xdeafbeef: will work for both block_id and strings and whatever impls display

```suggestion
  pub fn file_name<S>(block_id: S) -> PathBuf
    where
        S: Display,
    {
        PathBuf::from(block_id.to_string()).with_extension(Self::FILE_EXTENSION)
    }
```

core/src/storage/persistent_state/shard_state/writer.rs
```diff
@@ -459,3 +511,32 @@ enum CellWriterError {
...
+
+    pub fn finalize(self) -> ([u8; 32], W) {
+        (self.hasher.finalize().into(), self.inner)
+    }
+}
+
+impl<W: Write> Write for IntermediateHasher<W> {
+    fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
+        self.hasher.update(buf);
+        self.inner.write(buf)
```
0xdeafbeef: > Writes a buffer into this writer, returning how many bytes were written.
>This function will attempt to write the entire contents of buf, but the entire write might not succeed, or the write may also generate an error. Typically, a call to write represents one attempt to write to any wrapped object.

https://doc.rust-lang.org/std/io/trait.Write.html#tymethod.write


Underlying writer can write part of buf, but you assume that whole buf is always written.

hash only `&buf[..n]` returned from a write call.


It doesn't matter for `io::Cursor`, but may shoot in a foot for file.

collator/src/collator/do_collate/finalize.rs
```diff
@@ -100,7 +102,7 @@ impl Phase<FinalizeState> {
         let mut other_updated_top_shard_diffs_info = FastHashMap::default();
 
         for top_block_id in top_other_updated_shard_blocks_ids.iter() {
-            if top_block_id.seqno == 0 {
+            if top_block_id.seqno == zerostate_id.seqno {
```
0xdeafbeef: `top_block_id` be a shard block, so I'm not sure if this check is correct and what was the original idea.

I guess we should skip based on master ref seqno.

@SmaGMan please take a look

core/src/block_strider/starter/cold_boot.rs
```diff
@@ -38,22 +39,12 @@ impl StarterInner {
...
-                        "cold boot type cannot be changed after partial boot, \
-                        you must reset node state for that",
-                    );
-                }
-
+                tracing::info!("Starting node from genesis");
                 // Either import or download a zerostate.
                 let init_block = self.prepare_init_block(zerostates).await?;
-                assert_eq!(init_block.handle().id().seqno, 0);
+                // assert_eq!(init_block.handle().id().seqno, 0);
```
0xdeafbeef: drop commented out code

cli/src/cmd/tools/prepare_hardfork.rs
```diff
@@ -0,0 +1,314 @@
...
+            &BlockId {
+                shard: block_id.shard,
+                seqno: u32::MAX,
+                root_hash: block_id.root_hash,
+                file_hash: block_id.file_hash,
+            },
+            BlockMeta::with_data(NewBlockMeta {
+                is_key_block: false,
+                gen_utime: ssu.gen_utime,
+                ref_by_mc_seqno: ssu.min_ref_mc_seqno,
```
SmaGMan: `ref_by_mc_seqno` is a seqno of master block which will include current one. And it is not `min_ref_seqno`. But in this case it may not matter. Anyway better to fix not to keep incorrect usage of the field which may confuse others when they will look thru the code for example.

collator/src/manager/tests/manager_tests.rs
```diff
@@ -2159,6 +2160,11 @@ async fn test_queue_restore_on_sync() {
         .processed_to_stuff
         .set_processed_to(shard, test_adapter.last_mc_blocks.get(&1).unwrap());
 
+    let mc_block_id = BlockId {
```
SmaGMan: Suggesting to rename into `next_mc_block_id` to make usage more clear. And in the next places as well.

collator/src/internal_queue/queue.rs
```diff
@@ -314,7 +318,8 @@ where
 
             let diff = match diff {
                 // If top shard block changed and diff not found, then bail
-                None if *top_shard_block_changed && mc_block_id.seqno != 0 => {
+                // TODO: err if <
```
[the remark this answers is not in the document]
MrWad3r: It is already there

cli/src/cmd/tools/prepare_hardfork.rs
```diff
@@ -0,0 +1,318 @@
...
+            last_key_block: None,
+            block_create_stats: None,
+            global_balance: ssu.total_balance.clone(),
+        })?);
+
+        let mut builder = CellBuilder::new();
+        ssu.store_into(&mut builder, Cell::empty_context())?;
+        let updated_master_state = builder.build()?;
+
+        let clone = updated_master_state.clone();
```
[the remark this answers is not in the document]
SmaGMan: nit: You can make this way
```
let root_hash = *updated_master_state.repr_hash();
```

PR #980, opened 2025-12-17
feat(network): add optional bootstrap peers refill task

pashinov:
PR resolve the problem of looking for neighbours for nodes not from bootstrap peers when all neighbours were lost (for example: network shutdown more than 1h)

Yes

Added .dht.bootstrap_peers_refill_period: Option with default value None. Enable refilling bootstrap peers in the routing table.

None

### COMPATIBILITY

Full compatibility

Not Required

### PERFORMANCE IMPACT

No impact expected

### TESTS

No

### DIFF
--- a/network/src/dht/mod.rs
+++ b/network/src/dht/mod.rs
@@ -65,6 +65,12 @@ impl DhtClient {
         Ok(added)
     }
 
+    pub fn add_allow_outdated_peer(&self, peer: Arc<PeerInfo>) -> Result<bool> {
+        anyhow::ensure!(peer.verify(now_sec()), "invalid peer info");
+        let added = self.inner.add_allow_outdated_peer_info(&self.network, peer);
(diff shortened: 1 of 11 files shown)


network/src/dht/mod.rs
```diff
@@ -592,6 +607,35 @@ impl DhtInner {
...
     }
 
+    fn add_allow_outdated_peer_info(&self, network: &Network, peer_info: Arc<PeerInfo>) -> bool {
+        if peer_info.id == self.local_id {
+            return false;
+        }
+
+        let mut added = false;
+
+        let mut routing_table = self.routing_table.lock().unwrap();
```
0xdeafbeef: Scope the lock, holding it during `notify_waiters` can unnecessary contend lock

Rexagon (approved):

PR #984, opened 2025-12-19
rpc subscriptions

0xdeafbeef:


consensus/Cargo.toml
```diff
@@ -22,7 +22,6 @@ bitflags = { workspace = true }
 blake3 = { workspace = true }
 bumpalo = { workspace = true }
 bytes = { workspace = true }
-dashmap = { workspace = true }
 futures-util = { workspace = true }
```
Mododo: plz revert the removal, dashmap is widely used with a `FastDashMap` alias and there's no need to switch the import back and forth depending on whether it is currently imported or not (mostly for `dashmap::Entry`)

rpc-subscriptions/src/clients/id_pool.rs
```diff
@@ -0,0 +1,68 @@
...
+                    None
+                }
+            })
+            .ok()
+            .map(T::from)
+        {
+            Some(id) => AllocSource::Fresh(id),
+            None => AllocSource::Exhausted,
+        }
+    }
```
Mododo: looks like Ordering::Relaxed is enough for an atomic call without side effects

rpc-subscriptions/src/index/subscriptions.rs
```diff
@@ -0,0 +1,652 @@
...
+        for addr in addrs {
+            let addr: Addr = addr.into();
+
+            // Retries when the forward entry changes after we update the reverse index.
+            loop {
+                let Some((addr_id, created)) = self.ensure_addr_entry(addr) else {
+                    if let Some(ce) = entry.take()
+                        && ce.get().is_empty()
+                    {
+                        self.drop_reverse_entry(ce);
```
Mododo: looks like this part repeats `purge_client()`, so maybe remove `drop_reverse_entry()` and reuse `account_removed_reverse()` here instead?

rpc-subscriptions/src/memory.rs
```diff
@@ -0,0 +1,241 @@
...
+        needed.next_power_of_two()
+    }
+
+    #[cfg(test)]
+    pub(crate) fn len(&self) -> usize {
+        self.len.load(Ordering::Relaxed)
+    }
+}
+
+#[cfg(test)]
```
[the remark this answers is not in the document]
0xdeafbeef: done

Rexagon (approved):

PR #993, opened 2026-01-07
fix(storage): emit rocksdb metrics

0xdeafbeef:


storage/src/context.rs
```diff
@@ -180,7 +180,11 @@ impl StorageContext {
                     false
                 }
             }
-        }
+        };
+
+        // Store the updated items back
+        self.inner.rocksdb_instances.store(Arc::new(items));
```
Mododo: Looks like a `Mutex` is more appropriate than `ArcSwap` then.  
`rocksdb_instances` are read with `HashMap::get().cloned()` that's pretty fast.  
Currently we may miss some `name` during node init when multiple components call `add_rocksdb_instance()`

Mododo (approved):

PR #1062, opened 2026-04-01
feature(collator): remove dispatcher

drmick:
Replace the generic AsyncQueuedDispatcher abstraction (boxed dynamic closures, task descriptors, cancellation tokens) with a simple CollatorHandle using an enum-based command queue over mpsc channel


SmaGMan:
implemented in #1076

PR #1063, opened 2026-04-01
Feature/improved collation cancel

drmick:
- Removed CancelPending and simplified collator state management to Active/Waiting/Cancelled.
  - Added a new command_complete notify signal from collator to manager.
  - Implemented cancel_active_collators(...) in the manager: it marks active collators as cancelled, sends graceful cancel, and waits for the current
    command to finish.
  - Updated mismatch/sync paths to explicitly cancel active collators and wait before running sync.
  - Treated ExternalCancel as an internal graceful-cancel path (no regular on_cancelled dispatch handling).


SmaGMan:
implemented in #1083

PR #1065, opened 2026-04-09
feat(rpc): validate external message accept before broadcast

pashinov:


rpc/src/endpoint/proto/mod.rs
```diff
@@ -107,12 +107,21 @@ pub async fn route(State(state): State<RpcState>, Protobuf(req): Protobuf<Reques
...
+                        return ProtoErrorResponse {
+                            code: INVALID_BOC_CODE,
+                            message: e.to_string().into(),
+                        }
+                        .into_response();
+                    }
+                };
+
+                if let Err(e) = state.check_external_message(msg_cell).await {
+                    return error_to_response(e);
```
0xdeafbeef, line 123: Parse and validate must come as a single rayon spawn, otherwise spawn overhead can be >= check itself

pashinov: I’d keep it split for now. Small messages are parsed inline. Two Rayon tasks happen only for large BOCs where parsing/validation cost should dominate the spawn overhead. Also, there is get_account_state between parsing and validation. I’d rather not move shared state (read/locking) into the Rayon task just to save one spawn

Rexagon (approved):

PR #1067, opened 2026-04-09
mempool to slasher stats bridge

Mododo:
Model (pre-review): anchor stats belong to blocks in a slasher batch.
Corner cases:
* there is a `received_and_collated=true` fast path, when block is both collated and received from network. Then validation step is omitted and signatures are not reported to slasher. We actually have mempool stats for that block, but currently they are not reported.
* in case block collation was cancelled by a received block (`validate_block()` returns `ValidationStatus::Skipped`) anchor stats are collected but not reported to slasher too.

Slasher contract is left untouched intentionally until we agree on the models and code flow

The drawbacks of suggested design are:
* at every subset change mempool has to omit `max_consensus_lag_rounds`=210 rounds of anchor stats, otherwise old vset `validator_idx`s will be mixed with new one (mempool makes no difference between vset and subset changes), so subsets should not be changed too often
* stats attributed to anchor are older and don't belong to that anchor history: stats are for dag rounds GC'ed after that anchor and have at least `commit_history_rounds`=20 rounds lag

Inherited from general design: stats may not be reported to slasher near validation session end because detached validation tasks can finish after the session has already been closed on the slasher side; same for already flushed batches during the same session. Also in case a collated block is replaced by accepted by the quorum, its stats remain in slasher.

Parent branch code suggestion https://github.com/broxus/tycho/pull/1068: in `ValidatorSessionScope.remap_ids` use `PeerId` to `slot` map (instead of `validator_idx`->`slot`) to allow cross-subset stat collection in mempool (like `references_skipped`: see commit `fix(consensus): mempool stats across subset change`)

------------------------
### Updates after review

Model: anchor stats are imported during a slasher batch (disregarding block collation).

Old corner cases are removed.
New corner cases:
* collector batch was advanced only by validation progress (`on_block_validated()` or `on_block_skipped()`) - batches didn't overlap, so anchors imported for the first block were ignored
  * see `feat(slasher): batches overlap`; shouldn't we convert it into a PR to parent branch first, prior `TODO: Split or grow the previous batch to not discard events`?
* stats of anchors imported by `try_import_init_anchors()` are ignored
* `anchors_cache.remove_last_imported_above(next_chain_time)` removes anchors from block input, but stats are already reported - are anchors imported the same way on every node?


collator/src/collator/mod.rs
```diff
@@ -82,6 +83,7 @@ pub struct CollatorContext {
     pub config: Arc<CollatorConfig>,
     pub collation_session: Arc<CollationSessionInfo>,
     pub listener: Arc<dyn CollatorEventListener>,
+    pub recorder: Arc<dyn ValidatorEventsListener>,
```
SmaGMan: nit: may be rename into `stats_recorder`, or `stats_collector`, or `stats_listener`?

collator/src/manager/mod.rs
```diff
@@ -117,6 +118,7 @@ where
 
     collator_factory: CF,
     validator: Arc<V>,
+    recorder: Arc<dyn ValidatorEventsListener>,
```
SmaGMan: nit: may be rename into `stats_recorder`, or `stats_collector`, or `stats_listener`?

collator/src/collator/mod.rs
Mododo: fixed

collator/src/manager/mod.rs
Mododo: fixed

slasher/src/bc/mod.rs
```diff
@@ -159,16 +160,29 @@ pub struct SubmitBlocksBatch {
...
         let len = len.get() as usize;
 
+        let mut anchor_stats_history = map_ids
+            .iter()
+            .map(|validator_idx| AnchorStatsHistory {
+                validator_idx: *validator_idx,
+                stats: AnchorPeerStats { points_proven: 0 },
+            })
+            .collect::<Vec<_>>();
+        anchor_stats_history.sort_unstable_by_key(|a| a.validator_idx);
```
SmaGMan, line 181: `anchor_stats_history` is sorted here, but `signatures_history` is not sorted. But seems like it is required for them to be with the same sort order in the `TLRead::read_from` implementation

Mododo: That's right, because mempool emits stats already sorted and they are just passed as-is through collator.
Mempool stats don't need additional layer of `validator_idx`|`PeerId` remapping and sorting - there's a `PeerSchedule` inside mempool to keep all v(sub)set data

Mododo: misunderstood at first, fixed

PR #1071, opened 2026-04-17
fix mempool alt resolved metrics: fork vs late

### LINKED ISSUE #876: Fix `alt valid points resolved` metric
`tycho_mempool_points_resolved_ok` with `ord=alt` includes closed locations (`FirstResolved::Closed`), but there should be two distinct metrics for forks and late points.

mooncitydev:
## what this does

`tycho_mempool_points_resolved_ok` with `ord=alt` mixed together two different situations: equivocation (another digest already first resolved) and points that resolve after the inclusion location was already **closed** for signing. Issue #876.

This adds a `case` label only for **alt + valid** resolutions:

- `case=fork` — first resolved state is still `Valid` / `NotValid` for another digest (normal equivocation path).
- `case=late` — first resolved state is `Closed` (engine moved on / signing closed), so the valid point is not the same kind of fork signal.

First resolutions and error paths are unchanged (same labels as before).

## testing

Not run here (windows checkout of this repo fails on `test/data` paths with colons). CI should cover `consensus`.

built by mooncitydev

Fixes #876

### DIFF
--- a/consensus/src/dag/dag_location.rs
+++ b/consensus/src/dag/dag_location.rs
@@ -260,7 +260,17 @@ impl InclusionState {
                 resolved.alt(),
             )
         }
-        ValidateCtx::resolved(dag_point); // meter after checks
+        let alt_valid_case =
+            if matches!(dag_point, DagPoint::Valid(_)) && !dag_point.is_first_resolved() {
+                Some(match self.0.resolved.get() {
+                    Some(FirstResolved::Closed) => "late",
+                    Some(FirstResolved::Valid(_, _)) | Some(FirstResolved::NotValid(_)) => "fork",
+                    None => "fork",
+                })
+            } else {
+                None
+            };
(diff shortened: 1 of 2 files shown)


Mododo (changes requested):
A couple of things should be done to close the issue:
* fix line breaks
* add a third `ord`=`late` label value just for this case instead of a new `case` label, because late points aren't alternative point versions

Mododo:
Closed because of no activity for a month

mooncitydev:
you mean i have to stop working on contribution?

Mododo:
You haven't provided requested changes within a month. An issue should not create an impression of being in progress if it really isn't. You are welcome to create a new PR with appropriate fixes, we appreciate your interest in project. But I have to warn you, especially with the use of AI: low quality code will be rejected and accounted as an abuse.

PR #1074, opened 2026-04-28
RPC/Overlay rate limits

pashinov:


core/src/blockchain_rpc/rate_limits.rs
```diff
@@ -0,0 +1,131 @@
...
+use tycho_util::FastHashSet;
+use tycho_util::rate_limit::{RateLimitConfig, RateLimitPolicy, TokenBucketConfig};
+
+use crate::proto::blockchain::rpc;
+use crate::proto::overlay;
+
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+#[serde(default)]
+pub struct BlockchainRpcRateLimitsConfig {
+    limiter: RateLimitConfig,
```
Rexagon: Make all fields public, its config after all

core/src/blockchain_rpc/rate_limits.rs
```diff
@@ -0,0 +1,131 @@
...
+            traffic: config.traffic,
+            whitelist: config.whitelist.into_iter().collect(),
+        })
+    }
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(default)]
+pub struct BlockchainRpcTrafficLimits {
+    light_queries: TokenBucketConfig,
```
Rexagon: And `pub` here as well

core/src/blockchain_rpc/rate_limits.rs
```diff
@@ -0,0 +1,131 @@
...
+impl BlockchainRpcRateLimitPolicy {
+    fn classify(constructor: u32) -> BlockchainRpcTrafficClass {
+        match constructor {
+            overlay::Ping::TL_ID
+            | rpc::GetArchiveInfo::TL_ID
+            | rpc::GetPersistentShardStateInfo::TL_ID
+            | rpc::GetPersistentQueueStateInfo::TL_ID
+            | rpc::GetArchiveChunk::TL_ID
+            | rpc::GetBlockDataChunk::TL_ID
+            | rpc::GetPersistentQueueStateChunk::TL_ID
```
Rexagon: Persistent state/queue chunk reads probably should be treated as heavy queries

core/src/blockchain_rpc/service.rs
```diff
@@ -64,6 +65,11 @@ pub struct BlockchainRpcServiceConfig {
     /// Default: yes.
     pub serve_persistent_states: bool,
 
+    /// Rate limits for inbound blockchain-rpc traffic.
+    ///
+    /// Default: disabled.
```
Rexagon, line 70: But the default impl always enables rate limits

core/src/node/mod.rs
```diff
@@ -610,10 +610,18 @@ impl ConfiguredNetwork {
...
-            .named("blockchain_rpc")
-            .with_peer_resolver(self.peer_resolver.clone())
-            .build(blockchain_rpc_service.clone());
+        let public_overlay = {
+            let mut builder = PublicOverlay::builder(zerostate.compute_public_overlay_id())
+                .named("blockchain_rpc")
+                .with_peer_resolver(self.peer_resolver.clone());
+
+            if let Some(config) = base_config.blockchain_rpc_service.rate_limits.clone() {
+                builder = builder.with_rate_limiter(config.into());
```
Rexagon: nit: Accept optional rate limiter so that builder pattern is not split by if's?

network/src/overlay/rate_limits.rs
```diff
@@ -0,0 +1,97 @@
...
+#[derive(Clone)]
+pub struct PublicOverlayRateLimiter {
+    inner: Arc<dyn PublicOverlayRateLimitHandler>,
+}
+
+struct PolicyRateLimiter<P>
+where
+    P: PublicOverlayRateLimitPolicy,
+{
+    // TODO: normalize IPv6 -> /64
```
Rexagon: IPv6 will not work currently for `tycho-network` because it requires binding two endpoints on each ip type to communicate with other nodes

rpc/src/endpoint/jrpc/stream.rs
```diff
@@ -199,13 +206,28 @@ where
     let subs = Arc::<RpcSubscriptions>::from_ref(&state);
     let mc_rx = state.subscribe_mc_tick();
 
-    stream_route_inner(subs, mc_rx, params).await
+    let active_streams = req
```
Rexagon: Extensions can be passed via `Extension` extractor in arguments

rpc/src/endpoint/jrpc/stream.rs
```diff
@@ -199,13 +206,28 @@ where
...
+        .extensions()
+        .get::<ActiveStreamLimiter>()
+        .expect("ActiveStreamLimiter must be in extensions");
+
+    let ConnectInfo(addr) = req
```
Rexagon: Same here

rpc/src/endpoint/rate_limits.rs
```diff
@@ -0,0 +1,141 @@
...
+use serde::{Deserialize, Serialize};
+use tycho_util::FastDashSet;
+use tycho_util::rate_limit::{
+    RateLimitConfig, RateLimitPolicy, RateLimitVerdict, RateLimiter, TokenBucketConfig,
+};
+
+#[derive(Debug, Default, Eq, PartialEq, Clone, Serialize, Deserialize)]
+#[serde(default)]
+pub struct RpcRateLimitsConfig {
+    limiter: RateLimitConfig,
```
Rexagon: Make fields `pub`

rpc/src/endpoint/rate_limits.rs
```diff
@@ -0,0 +1,141 @@
...
+            limiter: RateLimiter::new(config.limiter),
+            traffic: config.traffic,
+        }
+    }
+}
+
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
+#[serde(default)]
+pub struct RpcTrafficLimits {
+    requests: TokenBucketConfig,
```
Rexagon: Make fields `pub`

util/src/rate_limit.rs
```diff
@@ -0,0 +1,238 @@
...
+pub struct RateLimiter<K, C>
+where
+    K: Clone + Eq + Hash + Send + Sync + 'static,
+    C: Copy + Eq + Hash + Send + Sync + 'static,
+{
+    inner: Arc<RateLimiterInner<K, C>>,
+}
+
+struct RateLimiterInner<K, C>
+where
```
Rexagon: Don't use `where` guards on structs, only on impls

util/src/rate_limit.rs
```diff
@@ -0,0 +1,238 @@
...
+            buckets: FastHashMap::default(),
+            rejects: 0,
+            cooldown_until: None,
+            last_seen: now,
+        }
+    }
+}
+
+struct TokenBucket {
+    rate_per_sec: f64,
```
Rexagon: Can we use fixed point math here? It is a hot path, so there are no reasons not to make it slightly faster

Rexagon (changes requested):

Rexagon (approved):

0xdeafbeef (changes requested):
I would drop stream rate-limits. Rpc is either behind cloudflare proxy => ConnectInfo always receives cloudflare ip and eventually bans all clients. Or it's behind local proxy => same ban of all clients. You should either get real ip from headers or remove this functional, cause it does more harm than  good.

0xdeafbeef (approved):
lgtm

PR #1083, opened 2026-05-12
explicit cancel collation flow

SmaGMan:
Use explicit flow to cancel active collations on the block mismatch or when received a block from bc far ahead of last collated. Now we drain active collation tasks in a separate cancel task and do not relay on collator result cancellation using flags. Removed races when shard collator trying to collate next block even if cancellation requested before sync.

Collation manager code was splitted on several modules.

None

None
---

### COMPATIBILITY

Fully compatible

Not Required

---

### PERFORMANCE IMPACT

No impact expected

---

#### Unit Tests

No special coverage

#### Network Tests

No special coverage

#### Manual Tests

Manual tests used on devnet5
- transfers 30k
- externals 100k
- deploy 30kk
- transfers 30k
- transfers 50k
- dex 2000

---

**Notes/Additional Comments:**  

Better to review all commits except the last one separately because the last commit only moves working code from a single file to a several modules.

### DIFF
--- a/collator/src/collator/do_collate/execute.rs
+++ b/collator/src/collator/do_collate/execute.rs
@@ -9,7 +9,7 @@ use super::finalize::FinalizeState;
 use super::phase::{Phase, PhaseState};
 use super::work_units::PrepareMsgGroupsWu;
 use crate::collator::do_collate::work_units::ExecuteWu;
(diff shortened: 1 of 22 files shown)


Rexagon (approved):

0xdeafbeef (approved):
lgtm

PR #1084, opened 2026-05-15
perf(consensus): reduce mempool points RocksDB stalls

0xdeafbeef:
points-rocksdb-flush-stalls

 We waited for 2 memtables before flushing, while live cleanup forced manual flushes and produced smaller-than-needed SSTs. This wasted flush work and still left only 2 immutable memtables before  rocksdb stopped writes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4185,6 +4185,7 @@ dependencies = [
  "blake3",
  "bumpalo",
  "bytes",
+ "bytesize",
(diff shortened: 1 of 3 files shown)


0xdeafbeef:
points-rocksdb-flush-timeline-humane

Mododo (approved):

Rexagon (approved):

PR #1098, opened 2026-06-08
Slasher (part 1)

Rexagon:
This PR adds an initial slasher impl. It observes and analyzes blocks validation (a signature collection step after collation). It gives a rough idea of how well each node works. When vset changes each node analyzes each others reports and emits some metrics (stake punishment will be implemented in later PR's).

### How does it work

Each validator collects `BlocksBatch`es that contain a matrix of received signatures from other validator per each block.
image

These block batches are then sent to the Slasher Contract (`message_ttl` / `message_retry_interval`) and at the same time each validator parses all transactions from that contract. So each participant knows all block batches from everyone.

When vset changes an analyzer is triggered and all batches are reduced into a list of validator ids to accuse. For now, accusations are only visible on metric `tycho_slasher_accusations_total{pubkey}`.

### Analyzer steps:

> NOTE: It is executed independently by each participant with their local configs.

- It first checks that the validator set covers enough blocks and has enough validators (`vset_len_threshold`).
- Then it scans stored block batches and, for each observer, counts only blocks that reached the 2/3+1 **weight** threshold of valid signatures. Those blocks become the observer’s usable samples.
- For each observer, it measures how often every other validator produced valid signatures on those usable samples. It uses the median valid-signature rate as a baseline, scales it by `slow_node_factor`, and flags validators whose statistically adjusted upper-bound rate is still below that slow threshold. Accusation weights from observers are accumulated per each validator.
- A validator is considered as "indeed accused" if 2/3+1 observers (by **weight**) accused it, or if it produced too much malformed batches (`malformed_samples_threshold`).

### Breaking Changes

Validators subset computation now uses a `vset_switch_round` from `consensus_info`.

### Config Changes

A new section has been added:
```jsonc
{
  "slasher": {
    // TTL of messages to the slasher contract.
    "message_ttl": "30s",
    // Interval between message delivery attempts.
    "message_retry_interval": "1s",
    // Additional time to wait for the previous batch delivery.
    "prev_delivery_timeout": "5s",
    // Validator set round must contain at least this amount of blocks to run analyzer.
    "vset_len_threshold": 1000,
    // At least this number of block samples must be collected to accuse someone.
    "block_samples_threshold": 100,
    // At least this number of malformed batches must be collected to accuse someone.
    "malformed_samples_threshold": 5,
    // We treat the node as slow if its block rate is this times the median rate.
    "slow_node_factor": 0.5
  }
}
```


collator/src/types.rs
```diff
@@ -354,39 +354,48 @@ pub(crate) type CollationSessionId = (ShardIdent, u32, u32);
...
     collators: ValidatorSubsetInfo,
     current_collator_keypair: Option<Arc<KeyPair>>,
 }
+
+// #[derive(Clone, PartialEq, Eq)]
+// pub struct SessionId {
+//     pub vset_switch_round: u32,
+//     pub catchain_seqno: u32,
+// }
+
```
0xdeafbeef, line 378: 🙄🙄🙄🙄🙄🙄🙄🙄🙄🙄🙄🙄🙄🙄🙄🙄🙄🙄🙄🙄🙄

Mododo (approved):

SmaGMan (approved):

PR #1108, opened 2026-06-29
fix(core): check root hash of downloaded zerostates

Rexagon:
Give a hard error when starting from an invalid zerostate.

Later we should move this check into the downloader itself, but there is no way of properly retrying the initial apply.

### DIFF
--- a/core/src/block_strider/starter/cold_boot.rs
+++ b/core/src/block_strider/starter/cold_boot.rs
@@ -476,13 +476,9 @@ impl StarterInner {
 
         let mc_block_id = self.zerostate.as_block_id();
         tracing::info!(%mc_block_id, "importing masterchain zerostate");
-        let root_hash = state_storage
-            .store_state_bytes(&mc_block_id, mc_zerostate)
+        state_storage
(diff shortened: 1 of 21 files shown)


core/src/block_strider/starter/cold_boot.rs
```diff
@@ -872,18 +869,31 @@ impl StarterInner {
...
+                Ok(root_hash) => {
+                    assert_eq!(
+                        &root_hash, expected_state_root,
+                        "storage must verify expected root hash"
+                    );
+                }
+                Err(e) => {
+                    tracing::error!("failed to store shard state file: {e:?}");
+                    remove_state_file().await;
+                    continue;
```
MrWad3r, line 894: Won't this `continue` create inconsistency after `apply_temp_cell` if any errors pop up?

MrWad3r (changes requested):

MrWad3r (approved):

MrWad3r:
lgtm apart from comment above. But after discussion we agreed it is intended

SmaGMan (approved):
lgtm

Mododo (approved):

PR #1100, opened 2026-06-17
Split persistent state on parts by shards

SmaGMan:
Introduces the split of the persistent state on to `2^persistent_state_split_depth files` by shards. It supports write/import from splitted files, exchange via RPC, backward compatibility with a single-file persistent state.

It writes persistent state metadata file indicating the split config even when no split required.
```
<block_id>.meta.json
---
{
  "split_depth": 2,
  "parts": ["0xa000000000000000", "0x2000000000000000"]
}
```

Final persistent state bundle will be:
```
<block_id>_part_0xa000000000000000.boc
<block_id>_part_0x2000000000000000.boc
<block_id>.boc
<block_id>.meta.json
```

If state does not contain accounts in some shard, no part file is generated for that shard.

Crash recovery:
- write sequence: parts, metadata, main file
- if main file does not exist, bundle is not finished
- already written persistent state parts will be reused on the next attempt 

### NODE CONFIGURATION MODEL CHANGES

Added optional param. Default = 0
```
{
  "core_storage": {
    "persistent_state_split_depth": 2
  }
}
```

### BLOCKCHAIN CONFIGURATION MODEL CHANGES

[None]

---

### COMPATIBILITY

Fully compatible. Existing single-file persistent state treated as created with `persistent_state_split_depth = 0`.

### SPECIAL DEPLOYMENT ACTIONS

[Not Required]

---

### PERFORMANCE IMPACT

State 10kk.

* persistent_state_split_depth = 2
```

write start: Jun 19 11:31
write stop: Jun 19 11:49
elapsed: 18 min
size: ~1.4G (each part by 350M)
```

* persistent_state_split_depth = 0
```
write start: Jun 19 11:30
write stop: Jun 19 12:42
elapsed: 72 min
size: ~1.4G
```

---

#### Unit Tests

Covered by:
- split_persistent_bundle_paths_are_rejected_for_zerostate_import
- persistent_state_meta_roundtrip
- shard_state_part_file_name_parser_recognizes_only_parts
- test_check_can_reuse_shard_state_part_files
- cache_rejects_split_sidecars_without_metadata
- persistent_state_writer_rejects_missing_absent_cell
- split_persistent_shard_state_import_from_dump
- overlay_server_split_persistent_state

#### Network Tests

[No coverage]

#### Manual Tests

Manual tests used
devnet5
- deploy 10kk
- init hack to write persistent on every key block on validator 2 and 3
- make a small change in blockchain config to force key block
- check persistent state creation

PR #1111, opened 2026-07-01
split persistent + treat raw import marker as a "poisoned state"

### LINKED ISSUE #1102: Use RawImportInProgress as a "poisoned" marker and require re-sync
Currently `cleanup_interrupted_raw_import()` removes only cells but does not update block handles and etc. Actually, the state is inconsistent even if `interrupted_raw_import` was cleared.

So we need to consider RawImportInProgress as a "poisoned" marker and crash node with a error that re-sync is required. In this case we need to set RawImportInProgress not only in `import_zerostates`, but in `download_shard_state` as well.

This is a follow-up for #1081 and #1100

SmaGMan:
Merge after #1100

---
Set `RawImportInProgress` marker on zerostate or persistent state import start and remove when import fully finished. If import fail the marker will persist. Node start will fail if `RawImportInProgress` marker exist and will require re-sync.

[None]

[None]

---

### COMPATIBILITY

Fully compatible

[Not Required]

---

### PERFORMANCE IMPACT

[No impact expected]

---

#### Unit Tests

Covered by:
- storage_open_rejects_existing_raw_import_marker
- finished_raw_import_marker_allows_storage_open

#### Network Tests

[No coverage]

#### Manual Tests

Manual tests used:
- start network
- deploy 1kk accounts
- force persistent state using HACK
- stop one node and drop its
- start node, wait when persistent import starts, then stop node
- try to start node - it should fail

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4574,8 +4574,7 @@ dependencies = [
 [[package]]
 name = "tycho-types"
 version = "0.3.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "104f8eaefe8d4c12e480906a876f0c6dd6f68c1e7eaeb8fd2c236e8f729019fd"
+source = "git+https://github.com/broxus/tycho-types.git?rev=b529060a4f229bc17f9a30cd12f90ee44931e16a#b529060a4f229bc17f9a30cd12f90ee44931e16a"
 dependencies = [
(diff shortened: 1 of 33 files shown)


Mododo:
lgtm

0xdeafbeef (approved):

Rexagon (approved):

PR #1112, opened 2026-07-01
fix(collator): fix validation for already skipped blocks

Rexagon:
There was a possibility that an already skipped block could be later used to start a validation task. It causes some races in tests and might as well waste resources on lagging nodes.


SmaGMan (approved):
lgtm

Mododo (approved):

PR #1114, opened 2026-07-10
fix(core): keep GC-protected block handles cached

0xdeafbeef:


core/src/storage/block_handle/mod.rs
```diff
@@ -358,11 +358,13 @@ impl BlockHandleStorage {
             let is_masterchain = block_id.is_masterchain();
 
             if block_id.seqno == 0
```
SmaGMan, line 360: Maybe compare with `zerostate_id`?

Rexagon: It is checking `IS_PERSISTENT | IS_ZEROSTATE` from meta now.

Rexagon (approved):

MrWad3r (approved):

